PR #1, opened 2022-09-20
Fix API Bug

hanywang2:
# Summary

This PR fixes the bug where API playgrounds aren't rendering properly from `<ParamField>` data

# Test Plan

- Just check that the code makes sense
- [Optional] double check Corrily's deployment and API playgrounds

### DIFF
--- a/client/src/utils/api.ts
+++ b/client/src/utils/api.ts
@@ -1,6 +1,7 @@
 import { ParamProps } from '@/components/Param';
 import { config } from '@/config';
 import { openApi } from '@/openapi';
+import { ApiComponent } from '@/ui/Api';
 import { AxiosRequestHeaders } from 'axios';
 import isAbsoluteUrl from 'is-absolute-url';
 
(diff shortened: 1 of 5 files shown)


client/src/utils/api.ts
```diff
@@ -177,13 +178,23 @@ export const getParamGroupsFromChildren = (children?: Children, auth?: string):
...
+  paramFields?.forEach((paramField) => {
     let paramType;
-    if (child?.props?.query) {
+    if (paramField.query) {
       paramType = 'query';
-    } else if (child?.props?.path) {
+    } else if (paramField.path) {
       paramType = 'path';
-    } else if (child?.props?.body) {
+    } else if (paramField.body) {
```
hahnbeelee, line 197: I have ptsd from conditionals failing when question marks are omitted. you might want to add them?

hahnbeelee (approved):
code lgtm as long as it works like it looks like it should and the preview comes out how it's supposed to.

PR #2, opened 2022-09-21
Automatically detect documentation framework when scraping

tobias-mintlify:
1. Detects which framework we are using based on specific tags each of them sets. You can manually specify the framework is the automatic detection fails.
2. Refactors the CLI to use folders.


hahnbeelee (approved):
lgtm! just ran all of ur gitbook, docusaurus, readme commands.

PR #5, opened 2022-09-23
Scrape GitBook with puppeteer

tobias-mintlify:
# Summary

GitBook nested navigation adds links with JavaScript. Images are also added in with JavaScript after the page loads.

Using Puppeteer lets us open all the menus by simulating clicking their toggles. Puppeteer also scrapes individual pages so we can wait for all network calls to finish before scraping -- allowing us to download the images.

# Test Plan

Run `scrape-page` on https://docs.gitbook.com/getting-started/gitbook-app

Run `scrape-section` on https://docs.gitbook.com/

You should see images downloaded when using both commands. You should also see pages in nested navigation scraped.

### DIFF
--- a/cli/package.json
+++ b/cli/package.json
@@ -21,7 +21,7 @@
     "inquirer": "^9.1.0",
     "minimist-lite": "^2.2.1",
     "node-html-markdown": "^1.2.0",
-    "puppeteer": "^17.1.2"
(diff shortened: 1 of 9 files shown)


hahnbeelee (approved):
LGTM! Images aren't working for me but nested navs are.

PR #6, opened 2022-09-24
Fix nested navigation showing up in the wrong anchors

tobias-mintlify:
# Summary

Root cause: Anchors were not searching for href recursively. I updated our functions to work recursively for arbitrary amounts of nesting.

# Test Plan

I tested this on Relate's docs. Further testing could try seeing how this works when multiple anchor's directories are in a single group.

For example:

```
{
 group: "Title",
 pages: [{
   group: "Title 2",
   pages: ["anchor1/page"]
 }, "anchor2/page"]
}
```

Should show up under anchor1 and anchor2 because the path on at least one page matches each.


client/scripts/local.js
```diff
@@ -56,19 +56,14 @@ const getFiles = async () => {
...
 
-        if (extension === "json" || extension === "yaml" || extension === "yml") {
+        if (extension === 'json' || extension === 'yaml' || extension === 'yml') {
           try {
-            outputFileSync(
-              "openapi",
-              Buffer.from(content, "base64").toString("utf-8")
-            );
-            const api = await SwaggerParser.validate("openapi");
-            console.log({api});
```
tobias-mintlify, line 66: Prettier wasn't running on all files. I fixed it, so expect to see lots of prettier changes like this file's for a few days.

hahnbeelee (approved):
LGTM!

hahnbeelee (changes requested):
JK NOT LGTM. Can you fix it so that the titles show up when it's nested and in a division?

tobias-mintlify:
I looked into it and the bug is occurring in `main` too. We should make a separate PR to address it.

PR #7, opened 2022-09-26
Integrate with Fathom analytics

tobias-mintlify:
# Summary

I made two major changes during this integration:

1. Switched from an interface to an abstract class for analytics implementations so they only implement the methods they need.
2. Added an `onRouteChange` method for analytics integrations that need to know when the route changes. PostHog will need to be updated to this method when we move it to dynamic imports.

# Test Plan

I'm adding Fathom's site ID to our mint config so we can run it in production for a while. Everything worked correctly locally.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 .DS_Store
+node_modules/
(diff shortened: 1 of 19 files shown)


hahnbeelee (approved):
I'm gonna pull a Han and say idk what's going on but I trust. :halfsalute:

PR #19, opened 2022-09-30
Add setup for endmatter

hanywang2:
# Summary
This PR adds the setup for "end-matter" (frontmatter but available at the end). It does the following:
- If there is frontmatter syntax at the end of the file, use that and remove it from view

# Test Plan
- All previews look normal for existing docs
- Title and description should be customized for Venice


hahnbeelee (approved):
LGTM

PR #24, opened 2022-10-02
GA4 Setup

tobias-mintlify:
# Summary

GA4 setup is different from other analytics integrations because we have to place a JS script instead of importing an npm package. The script is set conditionally to avoid unnecessary network requests for other users.

We still create an implementation object to send page views and custom events.

Docs are updated on how to create the streaming property GA4 needs.

I also fixed a bug in our other docs

# Test Plan

Added a test Google Analytics measurement ID to our `mint.json` so we can monitor the traffic for a week.

Use the preview links to see how the new Mintlify GA4 docs look.


hanywang2 (changes requested):
Build error

hanywang2 (approved):
Lgtm! Just make sure we remove the `>`s later

PR #25, opened 2022-10-03
Deprecate TipInfo and TipWarning

tobias-mintlify:
Deletes old callouts syntax.

Will merge after all client sites are building successfully.


hanywang2 (approved):
LGTM

PR #30, opened 2022-10-07
Validate URL before using it in the CLI

tobias-mintlify:
# Summary

Validates a link before attempting a network call.

# Test Plan

Run `npm run local` then try fake URLs. For example: `mintlify scrape-section www.nohttp.com`


hanywang2 (approved):
approved

PR #36, opened 2022-10-09
Generate nested navigation when scraping GitBook and ReadMe

tobias-mintlify:
# Summary

Navigation was flattened before in the `mint.json` the CLI generates. This correctly generates nested menus when they are being used.

# Test Plan

Run `mintlify scrape-section https://docs.readme.com/` and `mintlify scrape-section https://docs.gitbook.com/` in empty folders.

### DIFF
--- a/client/src/config.ts
+++ b/client/src/config.ts
@@ -101,12 +101,15 @@ export type Config = {
   };
 };
 
-export const findFirstPage = (group: Navigation, target: string): Page | undefined => {
+export const findFirstNavigationEntry = (
+  group: Navigation,
+  target: string
+): NavigationEntry | undefined => {
   return group.pages.find((page) => {
     if (typeof page === 'string') {
       return page.includes(target);
     } else {
-      return findFirstPage(page, target);
+      return findFirstNavigationEntry(page, target);
     }
   });
 };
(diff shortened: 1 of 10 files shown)


packages/mintlify/src/scraping/getSitemapLinks.ts
```diff
@@ -1,5 +1,7 @@
 import axios from "axios";
 
+// Not in use.
+// Gets all links in a sitemap.
```
hanywang2, line 4: If it's not in use, why is it not deleted?

hanywang2 (commented):
One comment. Other than that it lgtm

hanywang2 (changes requested):
Build failiures

hahnbeelee (approved):
lgtm!

client/src/config.ts
```diff
@@ -10,11 +10,11 @@ import configJSON from './config.json';
 
 export const config: Config = configJSON;
 
-export type Page = string | Navigation;
```
hahnbeelee, line 13: import Page

packages/mintlify/src/scraping/getSitemapLinks.ts
tobias-mintlify: It might be useful / needed in the near future. We can delete it in a few months if not.

PR #41, opened 2022-10-13
Fix CLI bugs from today

hahnbeelee:
## Summary
i made the silly little assumption that `yarn` is downloaded alongside `npm` (This is what Google and documentation told me) Testing in a guest account, I found out this was not the case. `yarn` was failing. This puts in checks to ensure that node_modules and yarn are downloaded.

### Notes
I created `mintlify clear`  and `mintlify install` for troubleshooting purposes.
Under the hood:
`mintlify clear` - runs `git clean -f` Git requires that you add the `-f` tag because there's no going back after running it. It's kinda a scary command so I took this out of the automation. For most of the customers' use cases, they won't be running multiple instances/pages like us so their files won't get bloated. On the other hand we should run this frequently. I can make this an automation for ourselves in the future...
`mintlify install` - basically just run `npm install` in the correct directory. This step is automated, but in the off chance (like today) that someone needs to run this manually it's good to have.

## Test Plan
If you want to be extra, make a guest account on your mac and test it. Otherwise I'm pretty confident this is a fix.


packages/mintlify/package.json
```diff
@@ -1,14 +1,14 @@
 {
   "name": "mintlify",
-  "version": "1.1.0",
+  "version": "1.1.01",
```
tobias-mintlify, line 3: 1.1.1 not 1.1.01

PR #43, opened 2022-10-13
Open all nested menus in Docusaurus before scraping

tobias-mintlify:
Docusaurus doesn't fetch all pages until you open the menus. This PR uses puppeteer to open each nested menu before scraping so all pages are in the HTML.

The PR also refactors Gitbook's puppeteer logic to share code with Docusaurus.


hahnbeelee (approved):
lgtm

PR #46, opened 2022-10-15
Openapi bugs

hahnbeelee:
## Summary
One of our customers' openapi specs don't have a schema so it wasn't working. This fixes it

## Test Plan (CRUCIAL)
Look at the deployments that have openapi specs defined and make sure they aren't affected.


tobias-mintlify (approved):
Preview links look like they're working

PR #54, opened 2022-10-25
[2/2] Feature(favicon): Add Favicon resources from the backend to be added to the frontend

hanywang2:
# Summary

This PR adds favicon resource URIs generated from the backend (see https://github.com/mintlify/server/pull/9) to be appended onto the frontend. It also removes former dependencies of the `favicon` package on the frontend and the `injectFavicon` function on prebuild


client/src/pages/_sites/[site]/[[...slug]].tsx
```diff
@@ -198,6 +254,7 @@ export const getStaticProps: GetStaticProps<PageProps, PathProps> = async ({ par
         config,
         openApi,
       }),
+      stringifiedFavicons: stringify(favicons),
```
Comment (hahnbeelee) [inline client/src/pages/_sites/[site]/[[...slug]].tsx:257]: if each url for a favicon is going to be following a certain format do we need to be sending favicon data?

Comment (hanywang2) [reply client/src/pages/_sites/[site]/[[...slug]].tsx:257]: I did it in case it'll change

client/src/pages/_sites/[site]/[[...slug]].tsx
```diff
@@ -65,12 +67,26 @@ interface ParsedDataProps {
...
   openApi?: any;
 }
+
+interface FaviconsProps {
+  'apple-touch-icon': string;
+  '32x32': string;
+  '16x16': string;
+  'shortcut-icon': string;
+  browserconfig: string;
+}
```
Comment (hahnbeelee) [inline client/src/pages/_sites/[site]/[[...slug]].tsx:77]: Did you cherry pick these particular types of files? A better thing to do might be to make `FaviconProps` type `{ rel: string, href: string, sizes: string }[]` and inside the `<Head>`:
```js
favicons.map((fav) => {
  return (
    <link rel={fav.rel} sizes={fav.sizes} href={fav.href} />
  )
}
```

client/src/utils/api.ts
```diff
@@ -58,16 +58,16 @@ const potentiallAddPathParams = (inputUrl: string, inputData: Record<string, any
...
-  if (contentType === 'multipart/form-data') {
-    let cleanedObj = removeEmpty(obj);
-    const bodyFormData = new FormData();
-    for (var key in cleanedObj) {
-      bodyFormData.append(key, cleanedObj[key]);
-    }
-
-    return bodyFormData;
-  }
+const getBody = (obj: Object, _: string) => {
```
hahnbeelee, line 61: what is this

hahnbeelee (changes requested):
Left some comments on both the front and backend. I think you either need to include all the different file types being generated on the frontend or stop generating them on the backend

client/src/utils/api.ts
hanywang2: A rebase from main

client/src/pages/_sites/[site]/[[...slug]].tsx
Comment (hanywang2) [reply client/src/pages/_sites/[site]/[[...slug]].tsx:77]: That's a good point

hahnbeelee (approved):

PR #55, opened 2022-10-29
Concurrency

hanywang2:
# Summary

This PR enables it so that `cacheAll` is ran on everytime `getPaths` is run. It's **really** important that `getPaths` only is run once during build and that it's only really called for the production instance.

# Test Plan

- Uncomment out the `process.env.NODE_ENV === 'production'` in path.ts. And refresh the page once.
- Check the output logs on the server to see that requests are being routinely made to check the status of the cache
- Once that cache completes, you should be able to run `redis-server` and `get [NAME OF ORG]` to see the cache of any deployment

### DIFF
--- a/client/next.config.js
+++ b/client/next.config.js
@@ -22,7 +22,8 @@ export default withSentryConfig(
     images: {
       disableStaticImages: true,
     },
-    webpack(config, options) {
+    staticPageGenerationTimeout: 1000,
+    webpack(config) {
       config.module.rules.push({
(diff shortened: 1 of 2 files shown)


hahnbeelee (approved):
Let's just merge this for now, but in order to make the dev experience better (and it's applicable to production. We need to refactor the backend.

I think we can just merge the cache-all and paths endpoint. There's no need for them to be separate. And we need to make it so that it only caches if the cache doesnt already exist. Then, during dev it wont have to cache everything every time.

PR #59, opened 2022-10-31
Generate dictionary of versions per href

tobias-mintlify:
# Summary

Generates a dictionary that tells you the version a page is in. We will use this to filter search results for versions not currently selected.

Algolia will still return pages in other versions so this is slightly inefficient, but a good 20/80 solution.

Actually filtering the search results will come in a separate PR.

# Test Plan

I fixed jest and added test cases for this function.


hahnbeelee (approved):
YES UNIT TESTS.

PR #62, opened 2022-11-02
Write Versioned Docs Docs

tobias-mintlify:
Docs for writing versioned documentation sites.

# Test Plan

Run Mintlify docs locally.


hanywang2 (changes requested):
go to Algolia and check that it's being index properly

PR #63, opened 2022-11-03
Enable Multiple OpenAPI specs for versioning

hanywang2:
# Summary

- Enables support for new type of openapi.json, which allows for multiple files
- Supports accompanying functions to support multiple openapi specs but also the original setup

# Test Plan

- Check that normal docs without openapi specs are good (builds)
- Check companies with openapi specs
- Check Dots's setup with multiple openapi specs


hahnbeelee (approved):
LGTM! I looked at the previews and they looked good - esp Dots.

PR #65, opened 2022-11-04
Auto generate RequestExamples without parameters

tobias-mintlify:
# Summary

Auto-generates cURL and Python request examples. The request example includes bearer and api-key headers as applicable.

screenshot of generated example

# Future Functionality
Include parameters in generated examples.
Fill-in bearer token and api-key with the user's input.

# Test Plan
See preview for Vital and Mintlify, and TwelveLabs.

Vital's is very useful because they use OpenAPI and already use RequestExamples -- so we should _not_ override their examples.

Mintlify shows a site without OpenAPI which will now have request examples generated.

TwelveLabs also uses OpenAPI but does not have prefilled request examples, so they are a good test that we still generate them when necessary.

### DIFF
--- a/client/src/utils/generateAPIExamples.tsx
+++ b/client/src/utils/generateAPIExamples.tsx
@@ -53,7 +53,7 @@ export function generateRequestExamples(
   const pythonSnippet = {
     filename: 'Python',
     code:
-      'import requests\n' +
+      'import requests\n\n' +
       `url = "${fullEndpoint}"\n` +
       `headers = {"accept": "application/json"${pythonAuthHeader}}\n` +
       `response = requests.${method?.toLowerCase()}(url, headers=headers)`,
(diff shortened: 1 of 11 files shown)


client/src/utils/generateAPIExamples.tsx
```diff
@@ -0,0 +1,90 @@
...
+  if (auth === 'bearer') {
+    pythonAuthHeader = `, "Authorization": "${authName ?? 'Bearer'} {YOUR_TOKEN}"'`;
+  } else if (auth === 'key') {
+    pythonAuthHeader = `, "${authName ?? 'key'}": "{YOUR_KEY}"'`;
+  }
+
+  const pythonSnippet = {
+    filename: 'Python',
+    code:
+      'import requests\n' +
```
hahnbeelee, line 56: im not biggest pythonista, but could you put another newline between the import and the rest of the snippet? I feel like it'll look better.

hahnbeelee (approved):

PR #71, opened 2022-11-08
Simple Array Types

hanywang2:
# Summary

Enable UI and functionality of sending simple array types

Screen Shot 2022-11-08 at 3 29 22 PM

# Test Plan

- Check that most docs work as normal and as expected
- Go to Aqueduct -> API Reference -> Account Owner -> Create a new account owner and test out various inputs for tags. Insert some data, and click on `Send Request`. You should see the right information in the Body data
Screen Shot 2022-11-08 at 3 31 44 PM


hahnbeelee:
One thing is that after I sent the request, all the other fields remained, but only the array input disappeared
image

hahnbeelee (approved):
LGTM except my one comment. Also one request - could you rename ApiPlayground back to Api? I think Pelm and UseKana are failing bc of it 😭 edit - I can also do it after you merge this.

PR #76, opened 2022-11-13
Write separate docs per analytics integration

tobias-mintlify:
Uses a nested menu to show each integration.

Users were confused seeing a big list of options when they only wanted to add a single integration. Having separate pages also lets us add details on where to find the api keys for each integration.


docs/site-stats/set-up/amplitude.mdx
```diff
@@ -0,0 +1,25 @@
+---
+title: "Amplitude"
+---
+
+Add the following to your `mint.json` file to sent analytics to Amplitude.
```
hahnbeelee, line 5: send* analytics. This typo exists on every page.

hahnbeelee (approved):
lgtm except for the one typo

PR #82, opened 2022-11-29
[1/x] Add Initial Blog Mode

hanywang2:
# Summary

The below only are in effect if `mode: "true"`

Screen Shot 2022-11-29 at 3 07 54 PM
Screen Shot 2022-11-29 at 3 07 45 PM

# Test Plan

- Simple code glance
- Check screenshot looks good


client/src/ui/Blog.tsx
```diff
@@ -0,0 +1,35 @@
...
+    <img src={image} alt={name} className="mr-3 w-7 h-7 rounded-full bg-slate-50 dark:bg-slate-800" decoding="async" />
+    <div className="text-sm leading-4">
+      <div className="text-slate-700 dark:text-slate-300">{name}</div>
+    </div>
+  </div>
+}
+
+const SHARE_ICONS_CLASSNAMES = "h-5 fill-slate-400 dark:fill-slate-500 hover:fill-slate-500 dark:hover:fill-slate-400 cursor-pointer"
+
+export function BlogSidebar() {
```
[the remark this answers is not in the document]
hahnbeelee: Also confused between the difference between BlogNav and BlogSidebar

client/package.json
```diff
@@ -51,6 +51,7 @@
     "brotli-size": "^4.0.0",
     "clean-css": "^4.2.3",
     "clsx": "^1.1.1",
+    "date-and-time": "^2.4.1",
```
tobias-mintlify, line 54: Can you write tickets for how we will allow customizing the time zone. Right now I assume everything is using GMT which will slightly shift dates for some customers.

client/src/ui/MDXContentController/MDXContentController.tsx
```diff
@@ -73,14 +79,15 @@ export function MDXContentController({
     contentWidth = 'max-w-4xl';
   }
 
+  const isBlogMode = meta.mode === 'blog';
```
hahnbeelee, line 82: Add question mark: `meta?.mode`

client/src/ui/MDXContentController/PageHeader.tsx
```diff
@@ -1,26 +1,39 @@
...
-  title?: string;
-  description?: string;
   section: string;
+  meta: PageMetaTags;
 };
 
-export function PageHeader({ title, description, section }: PageHeaderProps) {
+export function PageHeader({ section, meta }: PageHeaderProps) {
+  const currentPath = useCurrentPath();
+  const title = meta.title || slugToTitle(currentPath)
```
hahnbeelee, line 15: makes sense moving this logic here 👍

client/src/ui/MDXContentController/PageHeader.tsx
```diff
@@ -1,26 +1,39 @@
...
             {section && (
               <p className="mb-2 text-sm leading-6 font-semibold text-primary dark:text-primary-light">
-                {section}
+                {isBlogMode ? createdDateReadable : section }
               </p>
             )}
           </div>
-          <UserFeedback title="" />
+          {
+            !isBlogMode && <UserFeedback />
```
tobias-mintlify, line 35: Please refactor this before it gets messy. We'll likely add other modes in the future and these if-flags will be hard to refactor later.

The entire blog variable should not exist in this component.

I would suggest taking UserFeedback and Date as props. You can dynamically choose to pass them in based on whether the blog property exists outside this component.

client/src/ui/MDXContentController/PageHeader.tsx
```diff
@@ -31,6 +44,15 @@ export function PageHeader({ title, description, section }: PageHeaderProps) {
       {description && (
         <p className="mt-2 text-lg text-slate-700 dark:text-slate-400">{description}</p>
       )}
+      {
+        isBlogMode && <div className="mt-4 flex space-x-5">
```
tobias-mintlify, line 48: I like this better on the right now that we have other buttons there.

client/src/ui/Blog.tsx
tobias-mintlify, line 12: Here is where I'd move the authors.

client/src/types/metadata.ts
```diff
@@ -19,6 +19,13 @@ export type PageMetaTags = {
   auth?: string;
   size?: string;
   version?: string;
+  mode?: string;
+  _context?: {
```
tobias-mintlify, line 23: Remember to delete this when we move the values into `metadata`

client/src/ui/MDXContentController/TableOfContents.tsx
```diff
@@ -28,7 +29,10 @@ export function TableOfContents({ tableOfContents, currentSection }: any) {
     (section: TableOfContentsSection) => section.children.length > 0
   );
 
-  // Hide table of contents when its empty
+  if (meta.mode === 'blog') {
+    return <BlogSidebar />
```
tobias-mintlify, line 33: Please move this to MDXContentController by adding an if-branch in this if-statement:

```
{!isWideSize &&
        (isApi || requestExample || responseExample ? (
          <ContentSideLayout sticky>
```

tobias-mintlify (changes requested):

hahnbeelee (approved):
I'm mainly confused by the naming conventions of some of these files, but I guess it doesn't really matter because the UI is all going to be abstracted away into mintlify/components anyway.

The UI looks good. I still think the "Join" button on the right needs to not be so dark in light mode.

image

client/src/types/MDXContentControllerTypes.ts
```diff
@@ -1,21 +0,0 @@
-export type Meta = {
```
hahnbeelee, line 1: Also i think we should keep all types in the types folder. Even though it may seem like we're only using it in 1 files, it snowballs really fast once another location does need the type :/ I think it's better practice this way.

client/src/ui/Blog.tsx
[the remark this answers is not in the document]
hanywang2: Fair enough on the difference. Renaming

client/package.json
hanywang2: Great question. Awhile back I built this scheduling app that had to deal with timezones and schedules. So if you actually create the date object (`new Date()`) on the client, the timezones are accounted for. The data object in this case is built on the client, and then used for date-and-time, so it will always be localized

client/src/ui/MDXContentController/PageHeader.tsx
hanywang2: My thoughts on why it's not on the right now:
- The goal is to keep the right-side as empty as possible because we want to optimize for people to read the content with as few distractions as possible
- The authors aren't as important to be constantly viewed (probably why Stripe made it go away even though they easily could've kept it there with absolute position https://stripe.com/blog/new-and-improved-stripe-terminal). I personally have the preference that the author should only be acknowledged, but not reminded cause it doesn't matter that much
- I compared it side-by-side, and on the main page feels better to me (hard to explain)

PR #110, opened 2022-12-16
Use blocking for font and fix dark mode

tobias-mintlify:
Blocking will require the font to load before the site can be displayed, preventing flashing when the correct font is loaded.

Dark mode bug was caused by us storing `isDarkMode` as a boolean but our script was using strings.

### DIFF
--- a/client/src/layouts/SupremePageLayout.tsx
+++ b/client/src/layouts/SupremePageLayout.tsx
@@ -98,7 +98,7 @@ export default function SupremePageLayout({
               dangerouslySetInnerHTML={{
                 __html: `
                 try {
-                  if (localStorage.isDarkMode === true || (${(
+                  if (localStorage.isDarkMode === 'true' || (${(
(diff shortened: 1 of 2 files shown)


client/src/layouts/SupremePageLayout.tsx
```diff
@@ -98,15 +98,13 @@ export default function SupremePageLayout({
               dangerouslySetInnerHTML={{
                 __html: `
                 try {
-                  if (localStorage.isDarkMode === 'dark' || (${(
+                  if (localStorage.isDarkMode === true || (${(
```
hahnbeelee, line 101: nit: pls just do `localStorage?.isDarkMode`

tobias-mintlify: This code is a raw string we inject into a script tag on the front-end. Even if local storage is missing it can't bring down the page building.

PR #117, opened 2022-12-27
Open Anchors in new tab, move them to Next 13 Link component

tobias-mintlify:
Open anchors in a new tab when they link externally. Accomplished by expanding the DynamicLink component to take in more props that it flows through to Link.


hanywang2 (approved):
LGTM!

PR #120, opened 2022-12-30
Hahnbee/min 1386 make reusable components work for multi-tenancy

hahnbeelee:
## Summary
Snippets are reusable components/chunks of text. In order to create a snippet, it must live in the `_snippets` folder. In order to reference a snippet in content you must follow the following syntax: `<Snippet file="path/filename.mdx" />`. There is no need to import the component/file. More details about the product/tech spec [here](https://docs.google.com/document/d/1bb0OsoJoWAplFvwgmdoe1yuL7RUXPOumDRbkAMHfhgg/edit?usp=sharing)

- Takes in a snippets array in getStaticProps
- If a  Snippet component (`<Snippet file="path/filename.mdx" />`) is found in the content, it replaces it with the proper snippet.
- `remarkMdxInjectSnippet.ts` is probably the golden standard for plugins now. It's the best written plugin thanks to [wooorm](https://github.com/wooorm) ref: https://github.com/mdx-js/mdx/discussions/2210

### To-Do's
- [ ] Debug hydration error from react email's snippet
- [ ] Create PR's for react email and elementary and migrate them to VP
- [ ] Reflect this set-up in single tenancy
- [ ] Write docs for snippet
- [ ] Update CLI

## Test Plan
1. Check out to the corresponding backend PR branch `hahnbee/min-1386-make-reusable-components-work-from`
2. Navigate to snippet-test.localhost:3000 and make sure it looks the same with https://react.email/docs/introduction You can see the pages that have snippets by looking at [the changes I made in the contents] (https://github.com/zenorocha/react-email/compare/main...hahnbeelee:react-email:main).
3. Navigate to elementary.localhost:3000 and make sure it looks the same with https://docs.elementary-data.com/introduction You can see the pages that have snippets by looking at [the changes I made in the contents](https://github.com/elementary-data/elementary/compare/master...hahnbeelee:elementary:master)


client/src/utils/mdx/createSnippetTreeMap.ts
```diff
@@ -0,0 +1,133 @@
...
+    });
+    map = {
+      ...map,
+      [snippet.snippetFileLocation]: tree,
+    };
+  } catch {
+    // TODO: Better error handling
+  }
+  return map;
+};
```
hahnbeelee, line 85: This is really the core of creating the `treeMap` but the only reason why there is much more logic surrounding it is because we want to resolve instances where snippets reference other snippets.

client/src/utils/mdx/createSnippetTreeMap.ts
```diff
@@ -0,0 +1,133 @@
...
+    {
+      ...snippet,
+      content: parsedContent,
+    },
+    treeMap
+  );
+};
+
+/*
+ * We order the snippets by increasing number of references to other snippets in an attempt to avoid
```
tobias-mintlify, line 102: Is this ordering guaranteed to work or are there edge cases we should be aware of?

client/src/utils/mdx/remark/remarkMdxInjectSnippets.ts
```diff
@@ -0,0 +1,40 @@
+import type { Root } from 'mdast';
+import { MdxJsxFlowElement } from 'mdast-util-mdx-jsx';
+import { removePosition } from 'unist-util-remove-position';
+import visit, { SKIP } from 'unist-util-visit';
+import { VFile } from 'vfile';
+
+/**
+ * @typedef {import('remark-mdx')}
```
tobias-mintlify, line 8: Is this supposed to be at the top of the file?

client/src/utils/mdx/createSnippetTreeMap.ts
hahnbeelee: 2 edge cases:
- If the lowest number of references in a snippet is greater than 0 then there is no "base case" and the snippets will never resolve. https://github.com/mintlify/mint/blob/cc2e3754a52b4db02d0fdf786dccef368624d37d/client/src/utils/mdx/createSnippetTreeMap.ts#L26-L34
- It's possible that a snippet references another snippet that has not been resolved yet. We will continue to try to resolve failed snippets after editing the tree (because it's possible that the snippet its referencing was just added to the map)
https://github.com/mintlify/mint/blob/cc2e3754a52b4db02d0fdf786dccef368624d37d/client/src/utils/mdx/createSnippetTreeMap.ts#L42-L59

PR #128, opened 2023-01-05
Hahnbee/min 1409 check apikey is set in frontend

hahnbeelee:
## Summary
Unfortunately, the data we were sending from our backend was not always the way we expected it to be. In order to prune out empty objects, we specifically check for specific properties/values.

## Test Plan
- [x] console.logged the output from `getAnalyticsConfig` and checked that it was the output I was expecting.
I specifically checked when:
- we injected the analytics
- `mintConfig.__injected` was an empty object
- when it was defined in `mintConfig.analytics`


client/src/utils/getAnalyticsConfig.ts
```diff
@@ -1,18 +1,58 @@
...
+    analyticsConfig = {
+      ...analyticsConfig,
+      ...constructAnalytics(config.analytics),
+    };
+  }
+  return analyticsConfig;
 }
+// Note: It's really bad we have to do this because the data we were sending was not clean.
+// We added this as a sanity check to ensure analytics work as expected.
+// We should investigate improving our mongoose schemas to make this unnecessary.
```
tobias-mintlify, line 21: Self-hosting users may make mistakes regardless of what our hosted-solution server does. This validation is valuable :)

tobias-mintlify (approved):

client/src/utils/getAnalyticsConfig.ts
hahnbeelee: true!

client/src/lib/local/utils.ts
```diff
@@ -82,6 +82,9 @@ function optionallyRemoveLeadingSlash(path: string) {
 
 export const getSnippets = async (): Promise<Snippet[]> => {
   const snippetPath = 'src/_props/_snippets';
+  if (!(await pathExists(snippetPath))) {
+    return [];
+  }
```
hahnbeelee, line 87: this is just a leftover change from when I migrated getStaticProps to be single tenant. Doesn't affect the rest of the codebase and can be ignored

PR #131, opened 2023-01-06
Add support for default

hanywang2:
# Summary

Support default types on ParamFields

# Test Plan

- Check out docs with default types (radiant.mintlify.app) as an example


tobias-mintlify (changes requested):
Need to move `default` logic to ApiPlayground in mint.

PR #139, opened 2023-01-06
Add question mark to options check

tobias-mintlify:
Options is an optional value. We should have always had a ? in this if-statement.


benjamin-shen (commented):
A few lines down in the same function there's also `options.version`

benjamin-shen (approved):

PR #140, opened 2023-01-07
Update docs

hanywang2:
## Summary

Update the DNS configuration image to match the content


benjamin-shen (approved):
lgtm. Could you
1. Use a more specific pr title
2. Rebase from main so the old commit doesn't show up as part of this diff

Also OOC, where do these asset files live?

PR #143, opened 2023-01-08
Use Bearer: AUTH syntax

tobias-mintlify:
## Summary

We were incorrectly using `Bearer: Value` instead of `Authorization: Bearer VALUE` syntax.

## Test Plan

Tested locally on PropelAuth's docs.


hanywang2 (approved):
LGTM

client/src/utils/apiExampleGeneration/generateApiRequestExamples.tsx
```diff
@@ -127,13 +127,24 @@ function assembleUserInputHeaders(
...
     for (const headerParam of params.Authorization) {
-      if (
-        apiPlaygroundInputs.Authorization &&
-        apiPlaygroundInputs.Authorization[headerParam.name]
-      ) {
-        headers.push([headerParam.name, apiPlaygroundInputs.Authorization[headerParam.name]]);
-      } else if (headerParam.required) {
-        headers.push([headerParam.name, 'AUTH_VALUE']);
+      const isDefined =
+        apiPlaygroundInputs.Authorization && apiPlaygroundInputs.Authorization[headerParam.name];
```
benjamin-shen, line 131: I think you should explicitly cast it to a boolean if you are using it as one just so there's no funky type behavior or later confusion in the code, eg.
```suggestion
        !!(apiPlaygroundInputs.Authorization && apiPlaygroundInputs.Authorization[headerParam.name]);
```

Also, can the variable name `isDefined` be a bit more specific?

PR #144, opened 2023-01-08
Scrape website logo

benjamin-shen:
## Summary
- Cleaned up some of the scraper function types
- Renamed navigation types and moved them to global scope
- Fixed `--overwrite` flags for images
- Scrape website logo
- Refactored some download image code

## Test Plan
The CLI should install logos to `images/logo/logo-light-mode.<ext>` for each supported doc framework. The `--overwrite` flag should work as well. I tested it on
- https://docusaurus.io/docs
- https://docs.gitbook.com
- https://docs.readme.com/main/docs/intro-to-readme


packages/mintlify/src/downloadImage.ts
```diff
@@ -33,3 +36,52 @@ export default async function downloadImage(
...
+
+export function isValidImageSrc(src: string) {
+  // We do not support downloading base64 in-line images.
+  if (src.startsWith("data:")) {
+    return false;
+  }
+
+  const imageHref = removeMetadataFromImageSrc(src);
+  if (!SUPPORTED_MEDIA_EXTENSIONS.includes(getFileExtension(imageHref))) {
+    console.error("Invalid image path " + imageHref);
```
tobias-mintlify, line 48: We should separate validation for getFileExtension from checking if the file extension is valid. Invalid file extensions should have a descriptive error message: "We do not support the file extension: EXTENSION".

Please also add a siren emoji to the start of the error messages: 🚨

packages/mintlify/src/scraping/site-scrapers/scrapeGitBookSection.ts
```diff
@@ -1,20 +1,28 @@
...
   cliDir: string,
+  imageBaseDir: string,
   overwrite: boolean,
   version: string | undefined
-) {
+): Promise<MintNavigationEntry[]> {
   const $ = cheerio.load(html);
 
+  // Download the logo
+  const logoSrc = $("img").first().attr("src");
```
tobias-mintlify, line 23: In case there are other `img` tags before the logo in the future, could we use this instead:

`$('div[data-testid="public.headerHomeLink"] img').first().attr("src")`

packages/mintlify/src/scraping/site-scrapers/scrapeDocusaurusSection.ts
```diff
@@ -1,24 +1,29 @@
...
   cliDir: string,
+  imageBaseDir: string,
   overwrite: boolean,
   version: string
-) {
+): Promise<MintNavigation[]> {
   const $ = cheerio.load(html);
 
+  // Download the logo
+  const logoSrc = $(".navbar__logo").children().first().attr("src");
```
tobias-mintlify, line 19: Instead of `children().first()` you can likely use `$(".navbar__logo img").first()`

packages/mintlify/src/downloadImage.ts
tobias-mintlify: By the way, we need to block downloads of images over 5Mb. If you can think of a fast way to add that check to this function please do so. :)

packages/mintlify/src/scraping/downloadLogoImage.ts
```diff
@@ -0,0 +1,24 @@
...
+  imageBaseDir: string,
+  origin: string,
+  overwrite: boolean
+) {
+  if (!isValidImageSrc(imageSrc)) return;
+
+  const imageHref = cleanImageSrc(imageSrc, origin);
+
+  const ext = getFileExtension(removeMetadataFromImageSrc(imageSrc));
+  const imagePath = path.join(imageBaseDir, "logo", "logo-light-mode." + ext);
```
tobias-mintlify, line 21: Nice!

tobias-mintlify (changes requested):
Great work! I commented a few nits

packages/mintlify/src/downloadImage.ts
benjamin-shen: I researched a bit and didn't find a good way. We would probably need to download the image at least once, and then from there we could determine whether or not to save it (and possibly cache this information in a local file).

packages/mintlify/src/scraping/site-scrapers/scrapeGitBookSection.ts
benjamin-shen: Thanks for all the cheerio suggestions!

tobias-mintlify (approved):
Thank you for all the improvements you made alongside this ticket. :)

PR #147, opened 2023-01-10
Fix bug to support dynamic API paths

hanywang2:
## Summary

Fixes a bug that allows for the API Base to also be set by path parameters. In the following example, the `domain` can be set using a path param

```
---
api: "POST https://{domain}/api/backend/v1/magic_link"
---

<ParamField path="domain" type="string" required>
  The domain for your application
</ParamField>
```

CleanShot 2023-01-09 at 22 29 13@2x

### DIFF
--- a/client/src/utils/api.ts
+++ b/client/src/utils/api.ts
@@ -167,7 +167,7 @@ export const extractBaseAndPath = (
   }
 
   const url = new URL(fullEndpoint);
-  const base = url.origin;
+  const base = decodeURI(url.origin);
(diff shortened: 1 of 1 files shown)


hahnbeelee (approved):

client/src/utils/api.ts
```diff
@@ -167,7 +167,7 @@ export const extractBaseAndPath = (
   }
 
   const url = new URL(fullEndpoint);
-  const base = url.origin;
+  const base = decodeURI(url.origin);
```
benjamin-shen, line 170: decodeURI can throw an error, I think it might be worth including a try/catch here

PR #151, opened 2023-01-11
Run client inside Docker

tobias-mintlify:
## Summary

Our previous Docker file ran the old file listener code. Also, our middleware hardcoded localhost:3000.

This PR updates the code to run on the latest file listener code and successfully exposes the client to port 3333. We use 3333 because we do not support letting users customize the port yet and port 3000 is frequently used by other services.

We had to update the middleware to allow renaming any localhost with a four-digit port number.

## Other Small Changes

Started using `node:alpine` instead of the Node 19 image because it requires less memory.

Disabled Next.js telemetry when in Docker.

Moved `debugging-next.png` image out of the `client` folder so it doesn't get copied into Docker when we only use it in the README.

## Test Plan

Build: `docker build -t mint .`

Run inside the docs folder: `` docker run -it --rm -p 3333:3000 -v `pwd`:/app/user-working-directory mint ``

You should see the site at localhost:3333 rebuild as you make changes to the docs.

## Future Work

We can let the user customize the port in the future by changing the 3333 to any number the user inputs. Docker will always expose port 3000 and we simply map it to another port.


DOCKER_README.md
```diff
@@ -14,16 +14,12 @@ Building the docker image:
 
 ## Running the Mint Image
 
-Use this command to create and run a docker container with our image. The `--rm` flag deletes the container when we are done. Docker keeps all containers even after you disconnect if you don't add the `--rm` flag. Change 3020 to the port you want to run on:
+Use the command below command to create and run a docker container with our image in development mode. The `--rm` flag deletes the container when we are done. Docker keeps all containers even after you disconnect if you don't add the `--rm` flag. The `-v` flag mounts the folder where you run the command to `/app/user-working-directory` so Docker can read files as you edit them. Change 3333 to the port you want to run on.
```
benjamin-shen, line 17: "Use the command below command" => "Use the command below" or "Use the below command"

Dockerfile
```diff
@@ -15,12 +13,13 @@ COPY ./packages/file-listening ./file-listening
...
+
+# Expose the app's port.
 EXPOSE 3000
 ENV PORT 3000
 
-# By default, run the web service on container startup.
-# This command can be overriden by changing it in the docker run command.
-CMD ["sh", "-c", "yarn --cwd client run dev"]
\ No newline at end of file
+# The single & will run both commands in parallel.The double && is used for sequential execution.
```
benjamin-shen, line 23: Missing space
```suggestion
# The single & will run both commands in parallel. The double && is used for sequential execution.
```

DOCKER_README.md
benjamin-shen: You might also want to include an explanation for the `` `pwd`: `` part of the command. Also, does that syntax work on other operating systems too?

benjamin-shen (commented):
The code changes lgtm. I didn't go through the test plan yet

DOCKER_README.md
tobias-mintlify: Ah yeah it doesn't work on every operating system. We're testing tomorrow if the CLI will need to run it directly or if we can create a container in a more reliable way.

hahnbeelee (approved):

hahnbeelee:
Worked for me!

hahnbeelee:
loll the docker image shrunk in size 👀 
image

PR #153, opened 2023-01-13
Search single tenant

hanywang2:
## Summary

Remove search returning null when subdomain doesn't exist


client/src/ui/search/Search.tsx
```diff
@@ -257,7 +257,7 @@ export function SearchProvider({ subdomain, children }: { subdomain?: string; ch
   );
 
   const onSearch = async (query: string) => {
-    if (!query) {
+    if (!query || subdomain == null) {
```
tobias-mintlify, line 260: Maybe add a comment saying we don't allow search on local preview instances. No one will remember why we added subdomain == null in a year

hanywang2: Valid

tobias-mintlify (approved):

PR #162, opened 2023-01-13
Stop using structuredClone

hahnbeelee:
## Summary
[wooorm](https://github.com/mdx-js/mdx/discussions/2210) suggested we use structuredClone but in prod we're getting the error that it isn't defined. 
image

## Asks
I'm not sure why we would want it in the first place. Are there any down-sides to removing it? 🤔


hanywang2 (approved):

tobias-mintlify (changes requested):
Crashes because we are running on v16 when we should be in v18. I'm bumping Node to v18 in Vercel instead of merging this PR.

tobias-mintlify (approved):
Temporary

PR #163, opened 2023-01-13
Delete unused subdomain variable

tobias-mintlify:
We were getting a build warning. Other diff is prettier running.


hahnbeelee (approved):

benjamin-shen:
This is funny, it's the one warning that showed up from the barebones eslint config we were using before

PR #172, opened 2023-01-14
Setup format & lint for client directory

benjamin-shen:
## Summary
Sets up prettier and eslint for `client/`. I think we should only start to focus on `docs/` and `packages/` when we have a proper monorepo setup.


hahnbeelee (approved):
thanks for breaking my yarn :salute:

PR #173, opened 2023-01-14
Fix lint errors

hahnbeelee:
## Summary
Fix lint errors

## Test Plan
CI Check 👀


client/src/analytics/GA4Script.tsx
```diff
@@ -17,6 +17,7 @@ export default function GA4Script({ ga4 }: { ga4?: GoogleAnalyticsConfigInterfac
       />
       <Script
         strategy="afterInteractive"
+        id="ga4"
```
hanywang2, line 20: OOC is there a reason for this?

client/src/utils/api.ts
```diff
@@ -96,7 +96,8 @@ export const getApiContext = (
   if (inputData.Authorization) {
     const authEntires = Object.entries(inputData.Authorization);
     if (apiConfig?.auth?.method === 'basic' && authEntires.length === 2) {
-      let [[usernameField, username], [, password]]: any = authEntires;
+      const usernameField = authEntires[0][0];
+      let [[, username], [, password]] = authEntires;
```
hanywang2, line 100: Is this what ESLint is recommending? It looks harder to read

client/src/layouts/NavSidebar/BlogNav.tsx
```diff
@@ -1,7 +1,9 @@
+import Link from 'next/link';
+
 export function BlogNav() {
   return (
     <>
-      <a
+      <Link
```
hanywang2, line 6: I'm not 100% sure about this because I know Link tag has a weird behavior with classes.

But since blogs are insignificant, I wouldn't worry too much about it atm

hanywang2 (approved):
A few notes:
1. ESLint doesn't affect functionality 99% of the time, but there is a small handful of things in my exp where it has (changing a to Link or adding keys incorrectly) where it actually affects the system
2. Can we put some docs down on how to get everyone to install ESLint and auto-format on save properly? Just so that everyone has it

benjamin-shen:
@hanywang2 

> Can we put some docs down on how to get everyone to install ESLint and auto-format on save properly? Just so that everyone has it

Auto-formatting instructions are already in the readme, but I added some more info for linting. There are no extra steps to install ESLint, just `yarn` (same with Prettier).

client/src/analytics/GA4Script.tsx
benjamin-shen: > Error: `next/script` components with inline content must specify an `id` attribute. See: https://nextjs.org/docs/messages/inline-script-id  @next/next/inline-script-id

client/src/utils/api.ts
hahnbeelee: Because `usernameField` is not directly edited, it doesn't want it to be `let` so I moved it out to be `const` while keeping `username` and `password` `let`

hahnbeelee:
Can confirm what ben said, all i did was pull from main and everything was enabled already. If you want to see an overview of the errors you can run `yarn lint` in the terminal.

PR #175, opened 2023-01-14
[Validation] Add JSON Schema using zod-to-json-schema

hanywang2:
## Summary

Using the [zod-to-json-schema](https://www.npmjs.com/package/zod-to-json-schema) package, we're able to leverage and re-usage our existing validator setup with Zod to generate a JSON schema that can be added to our `mint.json` files

### Test

Added a test case, which passed so we should be good


hahnbeelee:
What are next steps? How do we plan on utilizing this?

packages/validation/package.json
```diff
@@ -14,7 +14,8 @@
   "scripts": {
     "build": "webpack",
     "test": "jest",
-    "publish-package": "npm run build && npm publish --access public"
+    "prepare": "npm run build",
+    "publish": "npm publish"
```
hahnbeelee, line 18: Did you add this so that it's easier to use with `yarn`?

packages/validation/tests/index.test.js
```diff
@@ -13,3 +14,9 @@ describe("validateMintConfig", () => {
     expect(results.errors.length).toEqual(1);
   });
 });
+
+describe("mintConfigSchema", () => {
+  test("zod to json schema generation is successful", () => {
+    expect(mintConfigSchema.$ref).toEqual("#/definitions/mintConfigSchema");
+  })
+})
```
hahnbeelee, line 22: Are there better ways to test the validation?

hanywang2: The schema is quite hard to read so I think the best way is to actually just have it being tested within the IDE. But it'll need to be set up in order for that to be done

packages/validation/package.json
hanywang2: Should I change it to `yarn build`? I'm just depreciate `publish-package` as with the other packages

hahnbeelee: Oh I see what you're doing. you should just remove the publish command, but keep the prepare command. The publish command is no different from just running `npm publish`

hahnbeelee (approved):
delete the publish command. other than that lgtm!

PR #176, opened 2023-01-14
Add option to hide api playground

benjamin-shen:
## Summary
Add to mint.json to hide API playgrounds:
```json
"api": {
  "hidePlayground": true
}
```

## Test Plan
Test this new mint.json option on any docs with API playgrounds.


packages/validation/src/schemas/config.ts
```diff
@@ -47,6 +47,12 @@ const apiSchema = z
...
         .optional(),
+      hidePlayground: z
+        .boolean({
+          invalid_type_error:
+            "hidePlayground must be a boolean. Try writing true or false without the quotes.",
+        })
+        .optional(),
     },
     {
       invalid_type_error:
```
tobias-mintlify, line 58: Also update the string down here to `api must be an object. The object can have baseUrl, auth, and hidePlayground as properties.`

benjamin-shen: nice catch

tobias-mintlify (approved):

PR #179, opened 2023-01-15
Refactor listener code

hahnbeelee:
## Summary
This should make it more stable.

We don't need to necessarily listen to `addDir` and `unlinkDir` commands because if you add/delete a folder with content it'll send `add` or `unlink` events and recursively add/delete the files. Although, if a user deletes a folder, an empty folder might remain. I will tackle this in a separate task that also handles deleting files when running the CLI in another directory.

## Test Plan
`mintlify dev` and create/delete files.
Edge cases
- [x] Snippets
- [x] Images
- [x] mint config
- [x] Add/delete folders


tobias-mintlify (approved):
Deleting a page you have open doesn't immediately redirect home (you need to refresh) but that seems acceptable.

PR #182, opened 2023-01-15
Download from a release tag

tobias-mintlify:
## Summary

Downloads GitHub repo as a tar file from the GitHub rest API instead of relying on users having Git installed locally.

This allows us to choose when we publish `mint` versions instead of always pulling from main.

## Test Plan

Ran locally using the `v.0.0.1` and `v0.0.2` releases.

You should be able to install and run this without changing anything in your `.mintlify` folder. We automatically delete the contents when the version changes. For testing, try changing the version in `mint-version.txt`, deleting the `client` folder, and running without internet.


benjamin-shen:
Thoughts on using a hidden file for `mint-version.txt`? Maybe `.mintrc` or `.mint`?

tobias-mintlify:
The folder `.mintlify` is already hidden.

tobias-mintlify:
Note this PR doesn't publish the CLI. We will bump and publish in a separate PR.

packages/mintlify/src/local-preview/index.ts
```diff
@@ -52,77 +54,103 @@ const dev = async () => {
...
+      // We do this because the folder name includes the commit sha, so we can't hardcode it.
+      // Lastly, we call .trim() to remove the newline character.
+      const extractedFolderName = shellExec(
+        'tar -tzf mint.tar.gz | head -1 | cut -f1 -d"/"'
+      ).stdout.trim();
+
+      // Delete tar file
+      fse.removeSync(TAR_PATH);
+
+      // Move extracted contents to CLIENT_PATH
```
hanywang2, line 114: Way too many comments. There's more comments in this particular function than lines of code.

If the comments just describe what the code is doing, delete it. It should only be used to supplement with context

hanywang2 (commented):
Delete some comments. There's way too much

packages/mintlify/src/local-preview/index.ts
hanywang2: https://stackoverflow.blog/2021/12/23/best-practices-for-writing-code-comments/

packages/mintlify/src/local-preview/index.ts
```diff
@@ -52,77 +54,103 @@ const dev = async () => {
...
-    if (firstInstallation) {
-      logger.succeed("Installation complete");
-    } else {
-      logger.succeed("Dependencies updated");
+
+    if (downloadTargetMint) {
+      // Delete any existing contents
+      fse.emptyDirSync(MINT_PATH);
+
+      logger.text = "Downloading Mint...";
```
hahnbeelee, line 86: The goal of the CLI isn't to let people know about the existence of mint. In fact we are purposefully trying to make mint accessible in a way in which you don't have to know about it.
```suggestion
      logger.text = "Downloading packages...";
```

packages/mintlify/src/local-preview/index.ts
```diff
@@ -52,77 +54,103 @@ const dev = async () => {
...
+
+      // Download the GitHub repo using the GitHub api
+      const octokit = new Octokit();
+      const downloadRes = await octokit.repos.downloadTarballArchive({
+        owner: "mintlify",
+        repo: "mint",
+        ref: TARGET_MINT_VERSION,
+      });
+
+      logger.text = "Extracting Mint...";
```
hahnbeelee, line 96:
```suggestion
      logger.text = "Extracting packages...";
```

hahnbeelee (changes requested):
Please update the messages! And ditto to what Han said, too many comments

hahnbeelee (approved):

PR #183, opened 2023-01-15
[Docs] Update Schema

hanywang2:
## Summary

Add `$schema` and remove `__injected` in the schema validations. `$schema` has been added as an actual (tho optional) config because I do believe it is needed to validate a `mint.json`

### Testing
- Tests have been added and is working


packages/validation/src/schemas/config.ts
```diff
@@ -150,6 +150,7 @@ const integrationsSchema = z.object(
 );
 
 export const configSchema = z.object({
+  $schema: z.string().optional().default('https://mintlify.com/schema.json'),
```
tobias-mintlify, line 153: Should do `z.string().url().optional().default(the url)`. Adding .url() is a shortcut for regex

packages/validation/src/index.ts
```diff
@@ -62,4 +62,8 @@ export function validateMintConfig(config: ConfigType): MintValidationResults {
   return results;
 }
 
-export const mintConfigSchema = zodToJsonSchema(configSchema, "mintConfigSchema");
\ No newline at end of file
+export const mintConfigSchema = (() => {
+  var schema = zodToJsonSchema(configSchema, "Schema") as any;
+  delete schema.definitions?.Schema?.properties.__injected;
```
tobias-mintlify, line 67: Can you delete `ultralight` and `ultradark` too?

tobias-mintlify (changes requested):

tobias-mintlify (approved):

PR #184, opened 2023-01-15
Wrap decodeURI in try-catch

tobias-mintlify:
decodeURI can throw if the URL is invalid. This has never happened to a customer but we are wrapping it anyway to prevent client-side exceptions we know can occur.


client/src/utils/api.ts
```diff
@@ -168,13 +168,20 @@ export const extractBaseAndPath = (
...
+    const base = decodeURI(url.origin);
+    const path = fullEndpoint.substring(fullEndpoint.indexOf(base) + base.length);
+    return {
+      path,
+      base,
+    };
+  } catch (e) {
+    return {
+      path: '',
+      base: '',
```
hanywang2, line 182: Idk if it'll work if path and base is empty. It'll likely throw an error down the line or make some weird UI

tobias-mintlify: The API call would fail, but it'd fail in the /api/ endpoint we use to send API requests which prevents the client-side exception

hahnbeelee, line 182: This doesn't seem like best practice. I haven't put a lot of thought into what would be best for this, but the expectation for functions should be that 
1. It takes in valid input (try to get the function inputs to be as consistent as possible)
2. Only return `null`/falsy values if it's an actual value
3. Throw errors otherwise

hahnbeelee (changes requested):
You probably weren't expecting this but lets see some better practices :)))

hahnbeelee (approved):

PR #189, opened 2023-01-17
Delete Docker MVP

tobias-mintlify:
We are not using Docker for the CLI. This PR deletes the duplicated code put in the file listener package and the Dockerfile.


hahnbeelee (approved):
Does a world exist in which we revisit this? What was your decision to not pursue this further?

tobias-mintlify:
We may revisit this one day. This PR will always be in our Git history to return to.

I'm deleting it because the file listener code it's using is already outdated so there's no point maintaining something we are not using.

hahnbeelee:
@tobias-mintlify Can you speak more about what went into not pursuing this further? What would revisiting this look like?

PR #190, opened 2023-01-17
Custom port numbers in the CLI

tobias-mintlify:
## Summary

Allows the user to pass a `--port` flag when running `mintlify dev` to pass a custom port. We do not automatically read env variables the user may have, so `PORT=3333 mintlify dev` wouldn't work.

## Test Plan

Tested the command worked on a variety of ports. Running on taken ports causes page loading to hang. For example, `mintlify dev --port 443` doesn't load on my computer because the port is used by HTTPS. Users customizing ports should be aware of this risk though.

I recommend using ports `3001`, `3002`, `3333`, and `3334` for testing.

## Future Work

We need to automatically switch ports when one is taken. We lost that functionality when we switched to `next-mdx-remote`. The package is not actively maintained so we may want to fork and tailor it to our needs. I'd also like to remove the part of the package that prints what port things are running on because we print that ourselves.


hahnbeelee (changes requested):
Update the README with information about defining the port!

hahnbeelee (approved):
🙌

PR #192, opened 2023-01-17
Terminate CLI port properly

hahnbeelee:
## Summary
Apparently using `spawn` is better for terminating ports properly than `exec` ([stackoverflow](https://stackoverflow.com/questions/20187184/how-to-kill-childprocess-in-nodejs)).

## Test Plan
1. Run `mintlify dev`
2. Terminate it
3. Run `lsof -i:3000`. There should be no node processes running. I think it's ok if the Google runs are running.
image


packages/mintlify/src/local-preview/index.ts
```diff
@@ -168,9 +169,28 @@ const run = (port: string) => {
...
+      PORT: port,
+    },
+    cwd: CLIENT_PATH,
+    stdio: "pipe",
+    shell: true,
+  });
+  mintlifyDevProcess.stdout.on("data", (data) => {
+    const output = data.toString();
+    console.log(output);
+    if (output.startsWith("> Ready on http://localhost:")) {
```
tobias-mintlify, line 184: Can we move these logs into here so they only show when the site is ready?

```
console.log(
    `🌿 ${Chalk.green(
      `Your local preview is available at http://localhost:${port}`
    )}`
  );
  console.log(
    `🌿 ${Chalk.green("Press Ctrl+C any time to stop the local preview.")}`
  );
```

tobias-mintlify: Also change `logger.succeed("Local Mintlify instance started.");` above to `Local Mintlify instance is ready. Launching your site now.`

And `Starting a local Mintlify instance...` to `Preparing local Mintlify instance.`

tobias-mintlify (changes requested):

packages/mintlify/src/local-preview/index.ts
```diff
@@ -151,26 +152,45 @@ const dev = async (argv: ArgumentsCamelCase) => {
...
```
benjamin-shen, line 176: This is kind of a sus check, is there some env variable or something we could check instead?

benjamin-shen (approved):

packages/mintlify/src/local-preview/index.ts
hahnbeelee: Oo maybe `lsof -i:3000` and check if it's running?

tobias-mintlify (approved):

PR #194, opened 2023-01-18
No local permanent redirect

hahnbeelee:
## Summary
When running mint locally, we shouldn't have permanent redirects because it'll permanently affect localhost:xxxx

## To-Do
@tobias-mintlify Can you make a new release of mint + the CLI? Btw did you write documentation for this? If not could you?

## Test Plan
👀


client/src/pages/[[...slug]].tsx
```diff
@@ -51,7 +51,7 @@ export const getStaticProps: GetStaticProps<PageProps, PathProps> = async ({ par
     if (Array.isArray(navWithMetadata) && navWithMetadata.length > 0) {
       const redirect = pickRedirect(navWithMetadata, slugStr);
       if (redirect) {
-        return redirect;
+        return { redirect: { destination: redirect.redirect.destination, permanent: false } };
```
Comment (benjamin-shen) [inline client/src/pages/[[...slug]].tsx:54]: I think you should destructure redirect all the way down to destination, and use the truthiness of destination as the if-guard. It'll look cleaner and we won't get any unexpected "property access of undefined" errors

client/src/pages/[[...slug]].tsx
```diff
@@ -51,7 +51,10 @@ export const getStaticProps: GetStaticProps<PageProps, PathProps> = async ({ par
     if (Array.isArray(navWithMetadata) && navWithMetadata.length > 0) {
       const redirect = pickRedirect(navWithMetadata, slugStr);
       if (redirect) {
-        return redirect;
+        const {
+          redirect: { destination },
```
Comment (tobias-mintlify) [inline client/src/pages/[[...slug]].tsx:55]: I think Ben meant to move that restructuring above the if-statement:

```
const {
   redirect: { destination },
} = redirect;
if (destination) {
   return { redirect: { destination, permanent: false } };
}
```

tobias-mintlify (approved):

PR #196, opened 2023-01-18
Support nested 3 layers deep

tobias-mintlify:
## Summary

Fixes: https://github.com/mindsdb/mindsdb/issues/4307

We were not defining padding for navigation 3-layers deep.

## Test Plan

Old:
old

New:
new


hanywang2 (approved):
LGTM

client/src/layouts/NavSidebar/DocNav.tsx
```diff
@@ -19,8 +19,10 @@ const getPaddingByLevel = (level: number) => {
       return 'pl-4';
     case 1:
       return 'pl-7';
-    default:
+    case 2:
```
hahnbeelee, line 22: I'm pretty sure the answer is no but is it not possible to do this programmatically? aka add integers instead of a switch case function?

hahnbeelee: O wait that would be so good actually

tobias-mintlify: I'll add a ticket

PR #197, opened 2023-01-18
Write Bash script to block deploys if client is not changed

tobias-mintlify:
Tested locally by running `bash shouldDeploy.sh` in the client folder. I updated the Vercel configuration to use the script but it may not take effect until the next deployment.


client/shouldDeploy.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+if [[ -z $(git diff --name-only HEAD~1 HEAD -- client) ]] ; then
+  echo "🛑 - Build cancelled. No changed to the client folder."
```
benjamin-shen, line 4:
```suggestion
  echo "🛑 - Build cancelled. No changes to the client folder."
```

client/shouldDeploy.sh
```diff
@@ -0,0 +1,17 @@
...
+  exit 0;
+fi
+
+if [[ "$VERCEL_ENV" == "production" ]] ; then
+  # Proceed with the build
```
benjamin-shen, line 9: I think this comment might be redundant since the message is self-explanatory

benjamin-shen (commented):
nice :)

hanywang2 (approved):
rubber stamped LGTM

PR #199, opened 2023-01-18
Validate version nesting

tobias-mintlify:
## Summary

Versions should never be set in groups nested in a group that already has a version. The page will render but the UI and navigation are unpredictable so we warn instead of erroring. For example, this is invalid:

```
{
  "group": "Group Name"
  "pages": [ { "version": "nested-version" } ]
  "version": "parent-version"
}
```

## Test Plan

Added unit tests. Can merge the PR if the tests pass.


packages/validation/tests/validateVersionsInNavigation.test.js
```diff
@@ -4,43 +4,95 @@ import {
...
+      ["v1", "v2", "v3"]
+    );
+    expect(results.errors.length).toEqual(0);
+    expect(results.warnings.length).toEqual(2);
+    expect(results.warnings[0]).toEqual(
+      "The version v2 is set on group Child Group that is nested inside a group with the version v1. Nesting versions is not supported and will cause unpredictable behaviour. Please do not set versions on groups nested inside a group that already has a version."
+    );
+    expect(results.warnings[1]).toEqual(
+      "The version v3 is set on group Child Child Group that is nested inside a group with the version v1. Nesting versions is not supported and will cause unpredictable behaviour. Please do not set versions on groups nested inside a group that already has a version."
+    );
```
hahnbeelee, line 94: I actually don't think this error message is very user friendly. It's a big information dump. I can see something like this being more easily consumable:

> Please do not set versions on groups nested inside a group that already has a version. Parent group "Parent Group" has version "v1" set. The nested group "Child Group" has version "v2" set. Please choose on or the other.

hahnbeelee (changes requested):
Improve warning message

hahnbeelee (approved):
ty <3

Issue #90, opened 2022-12-05
Structure of chapters is displayed differently than coded


hahnbeelee:
https://github.com/mindsdb/mindsdb/issues/4170

PR #200, opened 2023-01-18
Set padding based on header number instead of header order

tobias-mintlify:
## Summary

Setting based on header order never let us go back up a nesting level. The new approach has the limitation that you can't add a H4 header on its own without it being padded as if it were nested under H2 and H3, but that's questionable SEO and no one is doing that yet.

## Test Plan
tested


tobias-mintlify:
Fixes https://github.com/mintlify/mint/issues/90
Fixes https://github.com/mindsdb/mindsdb/issues/4170

hanywang2 (approved):

hahnbeelee (approved):

PR #201, opened 2023-01-19
Print validation messages when running locally

tobias-mintlify:
## Summary

Runs validation on mint.json in real time.

## Test Plan

Run in the docs folder and try breaking the JSON syntax then adding obviously wrong things like invalid HEX numbers or deleting the navigation.


packages/mintlify/src/local-preview/listener/index.ts
```diff
@@ -159,7 +161,44 @@ const onUpdateEvent = async (filename: string): Promise<FileCategory> => {
...
-      await fse.copy(filePath, targetPath);
+
+      const mintJsonFileContent = (await readFile(filePath)).toString();
+      try {
+        const mintConfig = JSON.parse(mintJsonFileContent);
+        const { status, errors, warnings } =
+          mintValidation.validateMintConfig(mintConfig);
+
+        errors.forEach((error) => {
+          console.error("🚨");
```
tobias-mintlify, line 172: I thought about making a function that takes care of printing the emojis but this code is easier to read.

packages/mintlify/src/local-preview/listener/index.ts
```diff
@@ -159,7 +161,44 @@ const onUpdateEvent = async (filename: string): Promise<FileCategory> => {
...
+          console.error(`🚨 ${Chalk.red(error)}`);
+          console.error("🚨");
```
hanywang2, line 174: Can we remove the emojis above and below the error itself?

hanywang2 (changes requested):

hanywang2 (approved):
LGTM

PR #202, opened 2023-01-19
Update to use relative path

tobias-mintlify:
The old command fails because it runs from `client` instead of from the root.

This is preventing Vercel from deploying our changes.


hahnbeelee (approved):
i was wondering if this worked lol. really no way to test this, plz observe post merge

PR #210, opened 2023-01-19
Fix home page overwriting folder redirects

tobias-mintlify:
## Summary

We were correctly finding redirects but were overwriting them with the home page redirect. This moves the logic into an else-branch and adds unit tests so this doesn't happen again.

I also removed a redundant if-statement in a child function -- that logic already existed higher up the code so it's safe to delete. The function also had unit tests that are still passing after deleting the check.

## Test Plan

Added unit tests. Safe to merge if the tests pass on the PR.


client/src/utils/staticProps/getFirstPageStartingWith.ts
```diff
@@ -29,15 +29,9 @@ const findFirstStartingWith = (entry: GroupPage, hrefStart: string): PageMetaTag
```
hahnbeelee, line 29: remove `isGroup` from imports since it isn't being used anymore

client/src/utils/staticProps/pickRedirect.ts
```diff
@@ -13,18 +13,20 @@ export function pickRedirect(
 } {
   let destination = null;
   let permanent = false;
+
   // Allow linking to a folder and redirecting to the first page in it.
   // Eg. The path "updates/changelog" can redirect to "updates/changelog/2022" but not "updates/changelog-page"
```
hahnbeelee, line 18: Can you move this description to a docstring for the function `getFirstPageStartingWith` ? That makes more sense to me.

hahnbeelee: Maybe you can rename the function `getFirstPageInDirectory`? I feel like that describes your example pretty clearly.

hahnbeelee: LOL my suggestion for the docstring actually reminds me of one of @benjamin-shen's PRs I reviewed for course plan. Check out `src/requirements/requirement-exam-mapping.ts` in [this PR](https://github.com/cornell-dti/course-plan/pull/606/files)

hahnbeelee (changes requested):
I wonder how the logic is affected if someone has folders in multiple anchors. For example, let's say I have an anchor for the `changelog` folder but higher up in my nav i have `changelog` used somewhere else too. Obviously the user should not be doing this. I wonder if the anchor will just not be accessible? probably lol

client/src/utils/staticProps/pickRedirect.ts
tobias-mintlify: This code doesn't affect anchors. Anchors use the config navigation but redirects use the PageMetaTags ones. That's why this code references `.href`.

I'm not moving the comment because the `getFirstPageStartingWith` we are using is very self-contained and can be used for more use cases than what we're doing with it here. `getFirstPageStartingWith` also doesn't have a concept of directories, it's looking for a exact string match hence the name.

hahnbeelee: This is a very general software engineering thing but I think it’s not the worst to make extremely purposeful functions that serve one function. In this case getFirstPageStartingWith is only being used for the purpose of serving essentially the functionality of your comment. And I would say this business implication is an important functionality. I kinda feel like creating a function like I described with a docstring with an example of that functionality is not the worse thing in the world for our use case considering it’s something more specific than pickRedirect too.

In the future I feel like we should do all path based things without caring about / aka handling slashes properly. None of this is important enough to refactor right now, but I just wanted to raise it for the future.

hahnbeelee (approved):

client/src/utils/staticProps/pickRedirect.ts
tobias-mintlify: We actually have a ticket to refactor some of the nav to not care about leading slashes! I was going to work on that this week.

PR #211, opened 2023-01-19
Simplify Auth code

tobias-mintlify:
## Summary

I was documenting authentication options and was confused how mint.json options overrode the page's own settings. Turns out the code is confusing! I've simplified it so instead of passing both page & mint.json settings, the auth function takes in the auth type and auth name. We figure out which config to use when we call the function instead of inside it.

Now, page `auth` properties override mint.json's auth property if you set both. No one was using both but this is a predictable matter we use in other parts of mint like when we override SEO tags or versions.

## Test Plan

Tested sites were the same using `yarn dev`.


hahnbeelee (approved):
Makes sense. Looks like an improvement from b4 🖖

PR #213, opened 2023-01-19
Add authentication page and make an API Playground nested group

tobias-mintlify:
## Summary

nav

## Test Plan

Documentation only changes. Site was tested with the CLI.


docs/api-components/examples.mdx
```diff
@@ -1,8 +1,10 @@
 ---
-title: "API Examples"
-description: "Display the request and response on the right sidebar"
+title: "Code in the Sidebar"
```
hanywang2, line 2: Happy to get some polling on this, but I think`API Examples` or another variation makes more sense because
1. It's always used for APIs. I don't think I've actually seen a time when it's not
2.  Sidebar is more often affiliated with navigation so it can be a bit misleading

hanywang2 (approved):

PR #218, opened 2023-01-20
Stop CLI if it's being ran in a directory without a mint.json

hahnbeelee:
## Summary
- Shows this error message when you run `mintlify dev` in a directory where a `mint.json` doesn't exist.
image

- Broke down the dev CLI scripts a tad so it's clear what each step is.
- Added the same check for when `yarn preconfigure` is ran
## Test Plan
Run `mintlify dev` in a directory with no mint.json


tobias-mintlify (approved):
Worked on my machine 🥳

PR #223, opened 2023-01-21
Handle gitbook code blocks properly

benjamin-shen:
## Summary
After this PR, the CLI scraper will handle Gitbook code blocks properly, including removing line numbers.

Also in this PR
- added @types/cheerio, which should help development with cheerio. this came with some type issues so I made some other minor fixes
- changed the gitbook header logic so headers weren't unnecessarily transformed into smaller headers

## Test Plan
```
mintlify scrape-page https://docs.gitbook.com/tour/editor/blocks/code-block
```
image

Note the existing bug documented below.

## Bug
Nested code blocks don't work; there is a parsing error. The html might look like this:
```
<pre><code>
  <div>outer</div>
  <div>```</div>
  <div>nested</div>
  <div>```</div>
  <div>outer</div>
</code></pre>
```

The translated code might look like this:
````
```
outer
```
nested
```
outer
```
````

The manual solution is to add an extra set of backticks to the outer code block. For example:
`````
````
outer
```
nested
```
outer
````
`````

This is an edge case (not exclusive to gitbook) that would probably require writing our own translator to fix.


tobias-mintlify (approved):
Great work! I'll publish the CLI immediately after this is merged.

PR #229, opened 2023-01-21
Improve auth method error message

tobias-mintlify:
## Summary

Old error message:

old-msg

New error message: `api.auth.method has to be one of: bearer, basic, key`

## Test Plan

Updated unit test to check the message.


hanywang2 (approved):
Looks good!

PR #237, opened 2023-01-23
Show 404 page on 403 error

tobias-mintlify:
We return 403 errors when a deployment doesn't exist for a page. Although an edge case, if it comes up again we want to show a 404 and log a warning instead of stopping the build step because the issue is not urgent.


client/src/lib/page.ts
```diff
@@ -16,8 +16,20 @@ export const getPage = async (subdomain: string, path: string) => {
...
+    }
+
     // Show a 404 page instead of crashing
-    if (axiosError?.response?.status === 400 || axiosError?.response?.status === 404) {
+    if (
+      axiosError?.response?.status === 400 ||
+      axiosError?.response?.status === 403 ||
+      axiosError?.response?.status === 404
+    ) {
       return { data: {}, status: axiosError?.response.status };
```
hahnbeelee, line 33: Shouldn't we send in `data: null` it seems like in getStaticProps we are not handling a 403 status.

hahnbeelee (changes requested):

client/src/lib/page.ts
```diff
@@ -16,8 +16,20 @@ export const getPage = async (subdomain: string, path: string) => {
...
     } else {
       throw error;
```
hahnbeelee, line 35: Do you know what would happen if we throw an error here? In `getStaticProps` should we be wrapping `getPage` in a try catch in the case that an error reaches here?

hahnbeelee (changes requested):

client/src/lib/page.ts
tobias-mintlify: Errors thrown prevents the project from advancing to production in Vercel. That's on purpose because we don't want to change the client when the server is returning 500s

client/src/lib/page.ts
tobias-mintlify: Good catch. Updated the if-statement in [[...slug]]

client/src/lib/page.ts
hahnbeelee: Are we sure this is the behavior?

tobias-mintlify: Yeah I saw it earlier before I deleted the pages. I'll DM you a link to the logs.

hahnbeelee (approved):

PR #242, opened 2023-01-24
Fix tar command on Windows by removing dependency on 'head' command

tobias-mintlify:
Windows doesn't have a `head` command which we used to find the name of the folder inside the tar file. Instead, we will use the tar command itself to remove the first directory.

## Test Plan

Tested tar command on Windows and tested `mintlify dev` on Mac. There may be other issues with other Windows versions but this fixes for at least some users.


hahnbeelee (approved):
On Mac OS:
- `mintlify dev` ran fine
- deleting `~/.mintlify` and running `mintlify dev` ran fine
- switching between different docs ran fine

PR #246, opened 2023-01-25
Remove use forward ref warning

NicoZweifel:
## Summary
Requires a new version of [`components`](https://github.com/mintlify/components) to be published first.
Should resolve ref warnings like `Function components cannot be given refs. Attempts to access this ref will fail. Did you mean to use React.forwardRef()?`
## Test Plan
Tested in https://github.com/mintlify/components/pull/31 but will require a new package version and possibly more changes before it is mergable.


client/package.json
```diff
@@ -7,8 +7,8 @@
     "build": "next build",
     "start": "next start",
     "export": "next build && next export",
-    "format": "prettier '**/*.{css,js,ts,tsx,md,mdx}' --write",
-    "format:check": "prettier '**/*.{css,js,ts,tsx,md,mdx}' --check",
+    "format": "prettier \"**/*.{css,js,ts,tsx,md,mdx}\" --write",
```
NicoZweifel, line 10: this will work on windows as well, while still working on unix.

tobias-mintlify (approved):
Worked when I tested it on the new package, thank you for this! I'll publish a new version of the Mintlify CLI with this change.

Issue #258, opened 2023-01-26
Hydration Errors (MDX plugin to move callouts out of p-tags)


NicoZweifel:
**Describe the bug**
Hydration error caused by invalid html.

**To Reproduce**
New next.js app will fail to hydrate when using `<div>` or any block element inside `<p>` when using MDX and code blocks.
In the mint client you can reproduce it with any text paragraph that contains code blocks for example something like this:
![image](https://user-images.githubusercontent.com/34443492/214959285-97e84075-ea08-4959-9863-4ec3a90abcca.png)

**Expected behavior**
~~When using code blocks like that `<p>` elements should be replaced with `<span>` when they contain child components and not just text.~~

**Screenshots**
![image](https://user-images.githubusercontent.com/34443492/214958322-cf7a5bec-9910-4361-baaf-31f1613e1712.png)

**Device:**
 - OS: Windows 10
 - Browser chrome
 - Version 109.0.5414.119 (Official Build) (64-bit)

**Additional context**
The Pop-ins that are happening with code blocks are not related to this and happen most likely because of hook usage.
You can verify this by replacing the MDX Components individually that are used for `code` and `p`.

I will create a draft PR to demonstrate.

NicoZweifel:
The issue seems to be resolved after [this ](https://github.com/mintlify/components/pull/39).
No more extra `<pre>` elements are  generated after the CodeBlock is correctly rendered/hydrated, so no replacing is neccessary anymore.

NicoZweifel:
After some more testing, noticed that the `<Note>`  Component also causes this error because it seems to use `<div>`. The Changes from earlier in #259  with changing to span if the content has child components fixes it.
I am not sure how to fix it otherwise, if we want to use components in the markdown directly between text:
```mdx
<Accordion title="Customize the branding">
    Branding mostly happens in the `mint.json` file. You can learn more about all available [configuration properties](https://mintlify.com/docs/settings/global).

    #### Update the logo

    Update `logo` folder with the light and dark version of your logo (can be svg or png).
    <Note>Make sure that the path matches the values set in `logo` of `mint.json`</Note>

    #### Customize the colors

    Populate the `colors` object with colors that match your brand

    #### Update the favicon

    Change the `favicon.png` with an image of the favicon

    #### Update footer social icons

    Update `footerSocials` with the right `[platform]: [url]` key-value pair
  </Accordion>```

tobias-mintlify:
Two ideas:

[...]

PR #259, opened 2023-01-26
fix hydration error

NicoZweifel:
## Summary
Fixes Hydration errors described in #258

This is just to demonstrate for now and should probably be improved a bit more, especially the component.
Regarding the pop-in of code blocks, it is caused by hooks, you can check by replacing the `CodeBlock` with `CodeBlock: 'code'` in `\mint\client\src\components\index.ts` and the pop-in will disapear.

## Test Plan

Tested locally, could not find any visual changes. One issue can be missing bottom margin when using `<span>` instead of `<p>`, but it usually only happens for pure text and in this case `<p>` is still used.

I checked normal text, Accordions, Cards and CodeBlocks and couldn't find an issue there so far.


tobias-mintlify (commented):
The missing margin you encountered is because Tailwind Typography, a plugin we use, only applies margins to headers and paragraphs. We also need to keep the parent `p` tag for accessibility purposes.

Do you have any sources I could read to learn why MDX doesn't like other tags nested inside `p`? I generally understand we aren't supposed to nest things inside `p` but I thought `code` was one of the accepted tags similar to `span`.

NicoZweifel:
The error itself is just related to next.js throwing this error when there are invalid elements.

As far as I understand it is because of the `pre` blocks because they are block elements.
If we managed to only place `code` when there is already a `p` element the error will disapear.
Maybe this is a better solution but I am not sure right now how to go about that. Could try by messing with the MDXComponents in another way.

NicoZweifel:
found this: 
https://stackoverflow.com/questions/5371787/can-i-have-a-pre-tag-inside-a-p-tag-in-tumblr
https://webmasters.stackexchange.com/questions/18150/html-should-code-be-placed-within-a-paragraph

NicoZweifel:
It turns out that the fix from [here ](https://github.com/mintlify/components/pull/39) with the Code Block actually also fixes the Hydration Error, there is no more extra `<pre>` element generated now that it is beeing rendered on the server.

NicoZweifel:
Leaving this PR/Issue for when the new package version is released.

NicoZweifel:
have this result now, no more warnings/errors and no pop in, but the margin isn't consistent, so I don't think we can use this yet:
image

NicoZweifel:
The Error in the starter Template can be removed by adding a newline after the text blocks, which will cause the paragraph to close before the calloutTemplates. No errors/warnings like this anymore and no margin issues either on local setup with this starter template.
https://github.com/mintlify/starter/pull/2

tobias-mintlify (approved):
I'm going to merge because the PR is only adding a type, but hydration errors can still occur if the user adds callouts without a blank line right?

NicoZweifel:
Yea it can occur if a user writes any block component without leaving a blank line in between the text. But we can track that in the issue, closing is fine with me 👍

PR #275, opened 2023-01-28
Setup segment with tracking mediator

hanywang2:
## Summary

Sets up Segment using [Analytics.js](https://segment.com/docs/connections/sources/catalog/libraries/website/javascript). The setup piggy backs off of the existing Analytics Mediator infrastructure.

### Note

Subdomain is now passed to the analytics mediator as it's passed as an event property to segment for querying the data for the right customers.


client/src/analytics/internal/Segment.tsx
```diff
@@ -0,0 +1,26 @@
+import { AnalyticsBrowser } from '@segment/analytics-next';
+
+const SEGMENT_WRITE_KEY = 'PqRXFGXckkpLI3dphj6IwgEGcrjSg83H';
```
hahnbeelee, line 3: Is this a security vulnerability now that this has been exposed? Should we make this an environment variable?

client/src/analytics/AnalyticsMediator.ts
```diff
@@ -38,9 +39,12 @@ export type AnalyticsMediatorConstructorInterface = {
 };
 
 export default class AnalyticsMediator implements AnalyticsMediatorInterface {
+  subdomain: string;
```
hahnbeelee, line 42: Now that we have both single and multi-tenancy, the `subdomain` keyword makes a lot less sense and I want to move towards using it less. Can we generically call this `id` or `siteId`?

client/src/analytics/internal/Segment.tsx
```diff
@@ -0,0 +1,26 @@
+import { AnalyticsBrowser } from '@segment/analytics-next';
```
hahnbeelee, line 1: Is there a reason why this is a `tsx` file? It doesn't seem like it uses JSX.

client/src/analytics/AnalyticsMediator.ts
hahnbeelee: jk scratch this for now. we can revisit this later...

client/src/analytics/internal/Segment.tsx
hahnbeelee: Apparently write keys are public so it's fine it's publicized. My main concern is how easy it is to rip this out for self service.

image
[Link](https://news.ycombinator.com/item?id=20886872)

hahnbeelee (changes requested):
My main concern is in regards to the isolation of this code. It seems like it's a layer added on top of the `AnalyticsMediator` because we want segment to run on every single event. A better way to construct this might be to add segment as another analytics implementation and have it always be "on" for us.

hahnbeelee (approved):
We have decided to move on for now for the sake of setting up segment and will return back to talking about the best object structure to implement for this.

PR #280, opened 2023-01-29
Segment implementation

hanywang2:
## Summary

A fairly sizable PR that does the following
1. Creates a Segment analytics setup for users to send events using their own Segment setups
2. Re-uses the Segment setup for our global/internal analytics + using an environment variable in getStaticProps
3. Updates the docs with the Segment implementation
CleanShot 2023-01-28 at 23 21 51@2x

### Note

- Make sure that `INTERNAL_ANALYTICS_WRITE_KEY` is set as environmental variables for all client deployments
- Need to publish validation package


client/src/analytics/AnalyticsMediator.ts
```diff
@@ -58,70 +66,89 @@ export default class AnalyticsMediator implements AnalyticsMediatorInterface {
       return;
     }
 
-    if (amplitudeEnabled) {
+    if (amplitudeEnabled && analytics.amplitude) {
```
tobias-mintlify, line 69: Not for this PR, but we should make a `validateIntegration` function that encapsulates these checks. There's others I'd like to add like making sure `analytics.amplitude` is not an empty object but I don't want the code getting messy. It'd be nice if we had a function that took a parameter name and did all the validation.

client/src/analytics/internal/index.ts
```diff
@@ -0,0 +1,5 @@
+import SegmentAnalytics from '../implementations/segment';
+
+const InternalAnalytics = SegmentAnalytics;
```
tobias-mintlify, line 3: Is this necessary?

client/src/analytics/AnalyticsMediator.ts
```diff
@@ -58,70 +66,89 @@ export default class AnalyticsMediator implements AnalyticsMediatorInterface {
...
       const posthog = new PostHogAnalytics();
-      posthog.init(analytics.posthog!);
+      posthog.init(analytics.posthog);
       this.analyticsIntegrations.push(posthog);
     }
+
+    try {
+      const internalTracking = new InternalAnalytics();
+      internalTracking.init({ writeKey: internalAnalyticsWriteKey });
+      this.internalTracking = internalTracking;
```
tobias-mintlify, line 120: You don't need this hardcoded. Try this instead:

```
if (internalAnalyticsWriteKey) {
   try {
      const internalTracking = new InternalAnalytics();
      internalTracking.init({ writeKey: internalAnalyticsWriteKey });
      this.analyticsIntegrations.push(internalTracking);
   } catch {
      console.warn('Invalid internal Segment write key');
   }
 }
```

`this.analyticsIntegrations.push(internalTracking);` has the same effect with less code. Delete the code below that creates an event listener, that'll automatically happen when the integration is in the array.

tobias-mintlify (changes requested):
You do not actually support Segment yet. You hardcoded the internal analytics use case and never check if the user set a write-key in mint.json. Your code should look like this pseudocode:

```
if (analytics?.segment?.writeKey) {
   const impl = SegmentImplementation(analytics?.segment?.writeKey)
   integrations.push(impl)
}
if (internalKey) {
   const impl = SegmentImplementation(internalKey)
   integrations.push(impl)
}
```

Make sure to follow the same pattern as the other implementations. I don't see a reason why your code needs to hardcode a special case.

Lastly, please put the internal implementation at the top of the AnalyticsMediator. We want events sent to the internal analytics first and pushing to the array first accomplishes that.

client/src/analytics/internal/index.ts
hanywang2: My rationale was that it
1. Makes it clear in the AnalyticsMediator that we are doing internal analytics for every event
2. It keeps it separate in case it changes and evolves in the future

client/src/analytics/AnalyticsMediator.ts
```diff
@@ -61,94 +66,77 @@ export default class AnalyticsMediator implements AnalyticsMediatorInterface {
...
     }
 
-    try {
-      const internalTracking = new InternalAnalytics();
-      internalTracking.init({ writeKey: internalAnalyticsWriteKey });
-      this.internalTracking = internalTracking;
-    } catch {
-      console.warn('Invalid internal Segment write key');
+    if (segmentEnabled) {
+      const posthog = new SegmentAnalytics();
```
tobias-mintlify, line 124: Typo. These should be called Segment not posthog

tobias-mintlify (approved):

PR #281, opened 2023-01-29
Add validation for static files

Graviis:
## Summary

Right now, Static files are not validated. This makes the user confusing as we are not showing any error and they expect the file to be upload. So from now on we will alert the user when the file is too big to be upload.


tobias-mintlify (approved):
This is really cool! The PR is going to help a lot of people thank you. :)

PR #282, opened 2023-01-29
Reduce z-index on nav bar

tobias-mintlify:
Part of: https://github.com/mintlify/components/issues/42

Going forward `z-50` should only be used for overlays. Having consistent z-indexes will make it easier to code overlays without knowing the exact index a component had.

## Testing

Tested sites on mobile and desktop. Also tested our various overlays were functioning, including the search bar.


hahnbeelee:
Is this the intended behavior?

image

tobias-mintlify:
No. The issue is being caused by backdrop blur. We can't fix the issue with backdrop blur until the z-index is fixed too though

hahnbeelee (approved):
all z-index related UI seemed normal to me

PR #283, opened 2023-01-29
Shorten size warning message to use relative paths

tobias-mintlify:
## Summary

The CLI warns when files are too big and won't be uploaded to Mintlify (over 5 Mb). This PR uses relative file paths to point out where the file is. The PR also moves the logic to make the file size check more modular.

## Test Plan

Tested locally by using a large video and seeing the CLI warned. Regular images do not cause warnings. Also verified the server allows anything `<= 5 Mb`.


tobias-mintlify:
I'm not sure how yet. I was looking into it but that code runs in a different function than where we run the check. We can try something in a separate PR.

hahnbeelee:
Can we make sure to refactor this so that it happens all in the same place?

hahnbeelee (approved):

tobias-mintlify:
@Graviis feel free to refactor anything if you're interested! The current design is ~ok and needs to be improved anyway

Issue #278, opened 2023-01-29
Scrape Intercom help centers


tobias-mintlify:
## Background

The Mintlify CLI has `scrape-page` and `scrape-section` commands to scrape other documentation websites into MDX format to help people migrate to the Mintlify platform. The commands automatically detect the platform and scrape the pages. Most of the logic to create a page is reusable but actually scraping each docs platform is different. You need to customize the scraping code to the HTML of the platform to know where nav links where and where the content itself is.

## Goal

Scrape Intercom help centres into `.mdx` files. For example: https://www.intercom.com/help/en/

## Tips

Break this up into three PRs.

1. Automatically detect whether a website is an Intercom help center. You will need to update `packages/mintlify/src/scraping/detectFramework.ts`. You can accomplish this by looking for `<meta name="intercom:trackingEvent` in the HTML.
2. Add `scrape-page` to download a single page's content into an MDX file.
3. Add scrape-section` that goes through the navigation to find all the pages and run `scrape-page` on each. Intercom collections should be treated as Mintlify groups when generating the `mint.json`.

See other scrapers for examples. There are helper functions you can use.

## Further Reading

[Cheerio, a library we use to help parse the HTML when scraping](https://cheerio.js.org/)

[Scraper docs](https://github.com/mintlify/mint/blob/main/packages/mintlify/scraper.md)

Graviis:
Im gonna take this one too ⚡️

Graviis:
@tobias-mintlify one question here, do we want to scrape the whole page? You pass: [Intercom](https://www.intercom.com/help/en/) or only sections? pass the link to one collection: eg: [Intercom section](https://www.intercom.com/help/en/collections/382-what-is-intercom)

tobias-mintlify:
Users would need to pass a link to the help center home page when running `scrape-section` so we can see all the collections. The scraper would need to open each section/collection and run recursively.

When scraping a single page with `scrape-page`, users should paste a link directly to the help center article they want.

PR #284, opened 2023-01-29
Scrape pages and sections from Intercom help center

Graviis:
## Summary

Add Intercom as a supported site to scrape. [Issue related](https://github.com/mintlify/mint/issues/278)


packages/mintlify/src/scraping/detectFramework.ts
```diff
@@ -37,5 +38,10 @@ export function detectFramework(html) {
     return { framework: Frameworks.README };
   }
 
+  const isIntercom = $("meta[name=intercom:trackingEvent]").length > 0;
```
tobias-mintlify, line 41: I think you need quotes inside this, eg: `'meta[name="intercom:trackingEvent"]'`

I can't test until later today though so feel free to comment if this is working

packages/mintlify/src/downloadImage.ts
```diff
@@ -49,6 +49,10 @@ export function isValidImageSrc(src: string) {
 
   const imageHref = removeMetadataFromImageSrc(src);
   const ext = getFileExtension(imageHref);
+
+  // TODO: We might need a different way of checking this because for example
+  //       against Intercom it's failing cause they use this types of urls
+  // https://downloads.intercomcdn.com/i/o/125665172/be372534f023143edd2f5caf/UBrQN_D_nTvozRUftyVHSw1eAc5aayle2-lBy9wuLpd4sqIfV31M5YFpE1pvQLvT7YQGZyFxjen07WY_DkYB41X3LfJCPACjhOI14SQxMqu3CPmp0I4XXdRgxytWnH0GbOSkMwWX
   if (!SUPPORTED_MEDIA_EXTENSIONS.includes(ext)) {
```
tobias-mintlify, line 56: Could we add an optional boolean parameter to this function that asks `validateImageExtensions`? Then the Intercom code can pass `false` when downloading all images and bypass this if-statement.

Also, we should pass `modifyFileName` when downloading images? We should split by `/` and only take the last part of the string. See GitBook's `scrapeGitBookPage` code for how its done there.

tobias-mintlify (changes requested):

packages/mintlify/src/downloadImage.ts
Graviis: yes, that will do!

packages/mintlify/src/scraping/detectFramework.ts
Graviis: it's working but im gonna add the quotes anyway

packages/mintlify/src/downloadImage.ts
Graviis: About this:
> Also, we should pass modifyFileName when downloading images? We should split by / and only take the last part of the string. See GitBook's scrapeGitBookPage code for how its done there.

Right now, without any modifications we are getting the last part of the string, so i think we are fine.

packages/mintlify/src/scraping/site-scrapers/Intercom/scrapeIntercomSection.ts
```diff
@@ -0,0 +1,54 @@
...
+      $ = cheerio.load(html);
+      const sectionTitle = $(".collection h1").first().text().trim();
+      const sectionPages = $(".section .g__space a")
+        .toArray()
+        .map((s: cheerio.Element) => $(s).attr("href"));
+      return {
+        group: sectionTitle,
+        pages: sectionPages,
+      };
+    });
```
Graviis, line 36: @tobias-mintlify  I have seen that for Docusaurus and Gitbook you do this in a separated function. do you think i should do the same for Intercom?

tobias-mintlify: You can keep it the way you have it. Docusaurus and Gitbook have a function because we reuse the code. I'll test the PR today thoroughly later today. :)

packages/mintlify/src/downloadImage.ts
tobias-mintlify: Do you know if these are always `jpeg` images? It would be good to add `.jpeg` to the end of it if so. Most users would not know how to open the images otherwise.

packages/mintlify/src/scraping/site-scrapers/Intercom/scrapeIntercomSection.ts
```diff
@@ -0,0 +1,54 @@
...
+    collections.map(async (entry: MintNavigationEntry) => {
+      return await scrapeGettingFileNameFromUrl(
+        entry,
+        cliDir,
+        origin,
+        overwrite,
+        scrapeIntercomPage,
+        false,
+        version,
+        "/docs"
```
tobias-mintlify, line 50: You don't need `/docs` here, that's used when all pages have a `/docs` directory at the start of the URL.

tobias-mintlify (approved):
Works great! While testing I noticed an improvement we can make elsewhere in the CLI:

Downloading images some times crashes if there are a lot of network calls made in parallel. Here's the error I got:

```
AxiosError: Client network socket disconnected before secure TLS connection was established
    at AxiosError.from 

... stack trace ...

    code: 'ECONNRESET',
    path: null,
    host: 'downloads.intercomcdn.com',
    port: 443,
    localAddress: undefined
  }
}
```

Maybe we could try-catch image downloading and log a friendlier error message when the error code is ECONNRESET? We could say `We encountered a network error trying to download from ${imagePath} please run scrape-section again.`

hahnbeelee:
@Graviis Can you add intercom as one of the frameworks in `packages/mintlify/scrapers.md`? And add any other relevant documentation where applicable?

Graviis:
> @Graviis Can you add intercom as one of the frameworks in `packages/mintlify/scrapers.md`? And add any other relevant documentation where applicable?

Sure!

Graviis:
> Works great! While testing I noticed an improvement we can make elsewhere in the CLI:
> 
> Downloading images some times crashes if there are a lot of network calls made in parallel. Here's the error I got:
> 
> ```
> AxiosError: Client network socket disconnected before secure TLS connection was established
>     at AxiosError.from 
> 
> ... stack trace ...
> 
>     code: 'ECONNRESET',
>     path: null,
>     host: 'downloads.intercomcdn.com',
>     port: 443,
>     localAddress: undefined
>   }
> }
> ```
> 
> Maybe we could try-catch image downloading and log a friendlier error message when the error code is ECONNRESET? We could say `We encountered a network error trying to download from ${imagePath} please run scrape-section again.`

Another solution could be open a browser and go through the pages instead of doing so many "unnatural" requests to the site

PR #286, opened 2023-01-29
[EZ] Remove newline for redeploy

hanywang2:
## Summary

Forgot to add the environment variable for analytics to work properly. Need to trigger redeploys to make sure it works.


hahnbeelee (approved):
We gotta figure out how to do this more efficiently

PR #287, opened 2023-01-29
Add initial pageload logging

hanywang2:
## Summary

The current setup only logs pageviews when the path is changed. This PR adds the setup to log on initial load.

Also fixes the inconsistent tooltip colors

### Testing

Add `console.log({ url, routeProps });` to `AnalyticsMediator.ts` and see that the initial page view is logged alongside all changes

```typescript
onRouteChange(url: string, routeProps: any) {
>  console.log({ url, routeProps });
  this.analyticsIntegrations.forEach((integration) => integration.onRouteChange(url, routeProps));
}
```
CleanShot 2023-01-29 at 13 05 47@2x


hahnbeelee (approved):
I'm realizing there is a lot of repeat code in the analytics mediator. We should consider refactoring it so we don't have to define whether it's in dev/prod in every init function.

PR #294, opened 2023-01-30
Generate canonical URL and og:url

tobias-mintlify:
## Summary

Canonical URLs tell search engines what a page is meant to be. For example, `/site?search=test` and `/site` should be the same site and Google should only link to the canonical `/site`.

Users do not pass their website domain to mint.json so we will generate the URLs using `window.location.origin`. There are disadvantages if a platform is scraping without JavaScript but this is an improvement until we have a more robust system. Future work could inject the URL when set on Mintlify's hosted platform. 

## Test Plan

Tested running locally uses `localhost` instead of a website's real URL because that's the origin set. Tested there were no hydration errors.


client/src/layouts/SupremePageLayout.tsx
```diff
@@ -49,7 +51,12 @@ export default function SupremePageLayout({
     };
   }, [navIsOpen]);
 
+  useEffect(() => {
+    setOrigin(window.location.origin);
+  }, []);
+
   const metaTagsDict = getAllMetaTags(pageMetadata, mintConfig);
+  const currentUrl = origin + useCurrentPath();
```
nawazdhandala, line 59: Why not directly do `window.location.origin + useCurrentPath();`

nawazdhandala (changes requested):

client/src/layouts/SupremePageLayout.tsx
naz-mintlify, line 59: Why not directly use `window.location.origin` instead of origin here? We can get rid of state,

naz-mintlify (changes requested):

client/src/layouts/SupremePageLayout.tsx
tobias-mintlify: The window property is not available when Next.js is pre-rendering the page on its server. Two errors can occur accessing window directly:

1. Window is undefined and the site won't build
2. If you use `typeof window !== "undefined"` to check if you're on the client, you would get a hydration error because the first render on the client has a different origin value than the server.

`useState` and `useEffect` let me pre-render without the origin and populate that with JavaScript immediately after the page loads. This will not work on search engines scraping without JavaScript.

Issue #260, opened 2023-01-27
Calculate nested padding programmatically instead of using Tailwind classes


tobias-mintlify:
Affects `client/src/layouts/NavSidebar/DocNav.tsx`

We are hard-coding several Tailwind padding levels. We should use `style={}` instead of `className=` and pass in the exact rem numbers we want. We can calculate the rem by multiplying the nesting level (minus 1) by the rem we're already using. See Tailwind's docs for the rem each class mapped to.

tobias-mintlify:
Thank you @naz-mintlify !

PR #295, opened 2023-01-30
Calculate left padding on navigation based on level

naz-mintlify:
## Summary

Issue: https://github.com/mintlify/mint/issues/260

Calculate left padding on navigation menu based on level. We were hard-coding several Tailwind padding levels. We now use style={} instead of className= and pass in the exact rem numbers we want. We calculate the rem by multiplying the nesting level.

## Test Plan

Tested running locally, create multi-level nav and with every level left padding should increase.


hahnbeelee (approved):
LGTM!

PR #297, opened 2023-01-30
Autogenerate meta images

tobias-mintlify:
Generates images using an API endpoint.

CleanShot 2023-02-03 at 22 07 17@2x

CleanShot 2023-02-03 at 22 06 01@2x

### Testing

- Use the `Meta Image Generation` endpoint in Postman.
- Generate an image with normal attributes filled
- Generate images with different values
- Generate an image with an extremely long description. It should be truncated


client/src/pages/api/og.tsx
```diff
@@ -0,0 +1,105 @@
...
+  try {
+    const interFontData = await interFont;
+    const { searchParams } = new URL(req.url);
+
+    const hasTitle = searchParams.has('title');
+    const title = hasTitle ? searchParams.get('title')?.slice(0, 100) : 'Powered by Mintlify';
+    const subdomain = searchParams.get('subdomain');
+
+    // Will crash when there is no logo for a subdomain and the subdomain was provided
+    const siteLogo = subdomain
```
hahnbeelee, line 32: Is there a way for us to check whether this image exists other than crashing?

client/src/pages/api/og.tsx
```diff
@@ -0,0 +1,105 @@
+// Private docs will need to update this function to prevent generating a meta image that exposes
+// title information to the public.
+// Docs: https://vercel.com/docs/concepts/functions/edge-functions/og-image-examples#encrypting-parameters
+
+/* eslint-disable @next/next/no-img-element */
+
+/* Disabled warning about using <img> instead of <Image> because this is an API route */
```
hahnbeelee, line 7: What happens when you use Image?

client/src/layouts/SupremePageLayout.tsx
```diff
@@ -55,7 +56,17 @@ export default function SupremePageLayout({
     setOrigin(window.location.origin);
   }, []);
 
-  const metaTagsDict = getAllMetaTags(pageMetadata, mintConfig);
+  // Fallback to Mintlify's hosted endpoint for search engines that don't support JavaScript.
+  const imageEndpoint = new URL(`${origin || 'https://mintlify.com/docs'}/api/og`);
```
hahnbeelee, line 60: Will `origin` still work for docs hosted in a subpath? for example pelm exists at pelm.com/docs will `/docs` be appended?

hahnbeelee, line 60: We should do some error handling for this because `new URL()` will throw if the URL is not a valid URL.

client/src/layouts/SupremePageLayout.tsx
```diff
@@ -55,7 +56,17 @@ export default function SupremePageLayout({
...
+  imageEndpoint.searchParams.append(
+    'title',
+    pageMetadata.title || pageMetadata.sidebarTitle || slugToTitle(pageMetadata.href || '')
+  );
```
hahnbeelee, line 64: Set `pageMetadata.title || pageMetadata.sidebarTitle || slugToTitle(pageMetadata.href || '')` to a const and check if it exists before appending it.

client/src/layouts/SupremePageLayout.tsx
```diff
@@ -55,7 +56,17 @@ export default function SupremePageLayout({
...
+  if (subdomain && mintConfig.logo) {
+    imageEndpoint.searchParams.append('subdomain', subdomain);
+  }
+
+  const metaTagsDict = getAllMetaTags(pageMetadata, mintConfig, imageEndpoint.toString());
```
hahnbeelee, line 69: Check this out - Ben sent it to me a couple days ago lmao. I think it'll help make this code more bulletproof

https://youtube.com/shorts/ps7EkRaRMzs?feature=share

client/src/utils/getAllMetaTags.ts
```diff
@@ -22,18 +23,18 @@ const SEO_META_TAGS = [
...
 
   const allMeta = {
     charset: 'utf-8',
     'og:type': 'website',
     'og:site_name': config.name,
+    'twitter:card': 'summary_large_image',
     'og:title': defaultTitle(pageMeta, config.name),
     'twitter:title': defaultTitle(pageMeta, config.name),
+    'og:image': imageEndpoint,
+    'twitter:image': imageEndpoint,
```
hahnbeelee, line 37: Are there other ones we should be adding this for?

hahnbeelee (changes requested):
Can you also add a screenshot of how this looks like?

hahnbeelee:
Can you also update the unit tests? / Look into that, they seem to be failing.

client/src/layouts/SupremePageLayout.tsx
tobias-mintlify: I'll use `set` instead of `append`, but that video is recommending to use URL like this code does haha

client/src/layouts/SupremePageLayout.tsx
tobias-mintlify: `slugToTitle(pageMetadata.href || '')` will return an empty string in the extreme case `pageMetadata.href` is not set and all other titles are not available

client/src/utils/getAllMetaTags.ts
tobias-mintlify: Not as far as brief googling suggests: https://css-tricks.com/essential-meta-tags-social-media/#aa-final-markup

client/src/layouts/SupremePageLayout.tsx
tobias-mintlify: We want this to throw an error because that would happen when `origin` is initialized to something other than an empty string, in which case this code is wrong and needs to be updated before deploying.

tobias-mintlify: `/docs` is not appended to the origin. That's part of the hardcoded Mintlify endpoint we fallback to.

client/src/pages/api/og.tsx
tobias-mintlify: I didn't try. I don't see a reason to because Next.js images are help with loading images after the site is loaded and/or to compress them but that doesn't exist in an api route.

tobias-mintlify:
@hahnbeelee Han is taking over this PR, description is intentionally left vague because this is not ready to merge yet. I should have written that in the description to begin with, sorry!

client/src/pages/api/og.tsx
tobias-mintlify: We might be able to make a `HEAD` request.

@hanywang2 please look into this when improve this endpoint

client/src/pages/api/og.tsx
benjamin-shen: Would it hurt? I'd rather stay consistent and minimize `eslint-disable`s if possible

tobias-mintlify: Even if it compiles I don't want to risk the unexpected behaviour. We want to send an `` tag so that's what we should be writing.

client/src/pages/api/og.tsx
hanywang2: This has been fixed by simply passing in the values of the image URIs directly

client/src/pages/api/og.tsx
```diff
@@ -0,0 +1,163 @@
...
+    const title = searchParams.get('title');
+    const description = searchParams.get('description');
+    const primaryColor = searchParams.get('primaryColor');
+    const lightColor = searchParams.get('lightColor');
+    const darkColor = searchParams.get('darkColor');
+    const isDark = searchParams.get('isDark');
+    const logoLight = searchParams.get('logoLight');
+    const logoDark = searchParams.get('logoDark');
+
+    const isDarkMode = Boolean(isDark);
```
benjamin-shen, line 33: This will make any truthy value of `isDark`, such as `isDark=false`, result in `isDarkMode=true`. Maybe you should use this?
```suggestion
    const isDarkMode = isDark === "true";
```

client/src/pages/api/og.tsx
```diff
@@ -0,0 +1,163 @@
...
+    const backgroundColor = isDarkMode ? '#0F1116' : '#FFF';
```
benjamin-shen, line 34: Do we want to use the user-defined background colors? Otherwise I'm worried that the other user-defined colors would clash or not be seen.

client/src/pages/api/og.tsx
```diff
@@ -0,0 +1,163 @@
...
+
+    const logo = isDark ? logoDark : logoLight;
+
+    return new ImageResponse(
+      (
+        <div
+          style={{
+            height: '100%',
+            width: '100%',
```
benjamin-shen, line 43: Is this necessary? Usually div width is 100% by default and [undesirable](https://www.impressivewebs.com/width-100-percent-css/)

client/src/pages/api/og.tsx
```diff
@@ -0,0 +1,163 @@
...
+            fontFamily: 'Inter var',
+          }}
+        >
+          <div tw="relative flex flex-col w-full px-26 pt-26 pb-28 h-full">
+            <svg
+              width="862"
+              height="450"
+              viewBox="0 0 862 450"
+              fill="none"
+              xmlns="http://www.w3.org/2000/svg"
```
benjamin-shen, line 56: What does this do?

benjamin-shen (commented):
IT'S BEAUTIFUL 💯

client/src/pages/api/og.tsx
tobias-mintlify: XML Namespaces define the SVG version you are using. It's only valid SVG if you include it. I doubt it'll impact rendering but this is a technically correct way to write the SVG.

client/src/utils/getOGImageEndpoint.ts
```diff
@@ -0,0 +1,46 @@
...
+  setQueryParamIfExists('lightColor', mintConfig.colors?.light);
+  setQueryParamIfExists('darkColor', mintConfig.colors?.dark);
+
+  if (typeof mintConfig.logo === 'string') {
+    setQueryParamIfExists('logo', mintConfig.logo);
+  }
+
+  if (mintConfig.modeToggle?.default === 'dark') {
+    imageEndpoint.searchParams.set('isDark', 'true');
+    setQueryParamIfExists('leftGradientColor', mintConfig.colors?.primary);
```
tobias-mintlify, line 29: You're already passing `primary` and `light`. Move this logic to the API endpoint.

client/src/utils/getOGImageEndpoint.ts
```diff
@@ -0,0 +1,46 @@
...
+    setQueryParamIfExists('rightGradientColor', mintConfig.colors?.light);
+
+    if (typeof mintConfig.logo !== 'string') {
+      setQueryParamIfExists('logo', mintConfig.logo?.dark);
+    }
+
+    return imageEndpoint.toString();
+  }
+
+  setQueryParamIfExists('leftGradientColor', mintConfig.colors?.dark);
```
tobias-mintlify, line 39: Same here, no need to pass this.

client/src/pages/api/og.tsx
hanywang2: Fair enough!

client/src/pages/api/og.tsx
hanywang2: Unfortunately in this case it is

hanywang2 (approved):

hahnbeelee (changes requested):
Hm when I navigate to http://mintlify.localhost:3000/quickstart I get this error. Note I already tried running `yarn`
image

hahnbeelee (commented):
Extremely long titles move the logo up and the description down
image
image

hahnbeelee:
From Han:
> I could implement something but for practical purposes it will never go over 3 lines and almost always never 2 lines

hahnbeelee (approved):

PR #299, opened 2023-01-31
remove z-index on parents/show tooltip in front of navbar

NicoZweifel:
## Summary
Fixes: https://github.com/mintlify/components/issues/42
Removes z-index on parent.
This is required, otherwise we create a new stacking context and cannot show elements (tooltips, popups) above the navbar.

## Test Plan

~~Requires package version to be bumped after https://github.com/mintlify/components/pull/48 is merged and published.~~
Tested locally with the mint docs.


tobias-mintlify (approved):
I'll approve if you want to merge in the WIP, but the tooltip is still showing in-front for me when running `yarn preconfigure` then `yarn dev-watch`. I'm on Chrome and MacOS.

branch

NicoZweifel:
I'm on Chrome on windows but I can check on Mac as well later today.
Maybe `yarn link` was still active? Could you try `unlink` and `yarn install --force` if you haven't already.
image

tobias-mintlify:
@NicoZweifel I just realized the tooltip works for me when the code block is in the page content like the screenshot you sent. The tooltip shows up behind the navbar when the code block is on the sticky right side bar like on http://localhost:3000/components/accordion

NicoZweifel:
Ah okay got it. Seems to be related to a new stacking context as well, I could not find a `z-index` on that element though, possibly because of `sticky`?

tobias-mintlify:
@NicoZweifel that's right! Removing `sticky` fixes it. It must be modifying something about the layering

tobias-mintlify (approved):
I'm going to merge this in because it fixes the issue for half the code blocks. Let's make the `sticky` fix in a separate PR.

PR #301, opened 2023-01-31
WIP: Make single tenant

naz-mintlify:


hahnbeelee (approved):
LGTM! Out of curiosity do you think we should require `IS_MULTI_TENANT` to equal `true` in order for it to work or do you think any value should work?

hahnbeelee:
Can you also update infisical with `IS_MULTI_TENANT` to true?

hahnbeelee:
Can we make sure to merge this when everyone is awake? It will impact all of our customers, we just want to make sure it goes smoothly.

naz-mintlify:
@hahnbeelee Updated infiscal. I believe any value should work. Users who self host will skip that env var altogether.

PR #306, opened 2023-02-01
Make github feedback links generate from mint.json

naz-mintlify:
## Summary

GitHub feedback links are now generated on the client vs the server.  The feedback links are generated from mint.json or are injected from the server if IS_MULTI_TENANT is set to true. 

## Test Plan

Please make sure you pull down the server and are on the same branch before you test this. 

Add GitHub config in mint.json and test IS_MULTI_TENANT = false first. Check if you see the feedback buttons and if they redirect properly. 

set github.isPrivate = false in mint.json and see if the buttons are hidden. 

Do the same test on IS_MULTI_TENANT = true and see if you see the button. Setting isPrivate = true in deployment.integrations will hide the buttons as well.


client/src/types/config.ts
```diff
@@ -88,6 +88,12 @@ export type Config = {
     icon?: string;
     color?: string;
   };
+  github?: {
+    org: string;
+    repo: string;
+    deployBranch: string;
+    path: string;
```
hahnbeelee, line 95: I think we briefly talked about this over a call, but make sure to set `isPrivate` too and if it is true, don't display the feedback buttons.

naz-mintlify: @hahnbeelee do we have isPrivate on the server as well in the `deployments.integration` collection?

naz-mintlify: I just added isPrivate

docs/mint.json
```diff
@@ -151,9 +145,16 @@
     "discord": "https://discord.gg/MPNgtSZkgK",
     "twitter": "https://twitter.com/mintlify"
   },
+  "github": {
```
tobias-mintlify, line 148: This is good for testing, but multitenant applications should inject this information from the server. The DB already has that information. You'll need to update the getStaticProps server endpoint and have it set `github` on mint.json  if it's not already set. That is, don't allow it to override anything the user specified in mint.json.

tobias-mintlify (changes requested):

docs/mint.json
naz-mintlify: I think I did that, should be a PR on the server. Would you like me to not hardcode mintlify values here?

hahnbeelee: Looks like this has been done in the backend PR?

naz-mintlify: Ah, yes. This was done on the backend. Let me revert and do it on client instead.

naz-mintlify: @hahnbeelee Yes!

client/src/ui/Feedback.tsx
```diff
@@ -43,85 +42,94 @@ const FeedbackTooltip = ({ message }: { message: string }) => {
...
-  const { createSuggestHref, createIssueHref } = useContext(FeedbackContext);
+
+  const [createIssueHref, setCreateIssueHref] = useState<string | null>(null);
+  const [createSuggestHref, setCreateSuggestHref] = useState<string | null>(null);
+
+  useEffect(() => {
+    if (path && mintConfig?.github && !mintConfig.github.isPrivate) {
+      const { owner, repo, deployBranch, path: deploymentPath } = mintConfig?.github;
+
+      // TO DO: We should add the / at the end when not set, but gracefully handle when users
```
tobias-mintlify, line 53: Please make a GitHub issue for this to do when you merge the PR

tobias-mintlify (approved):

docs/mint.json
hahnbeelee: @tobias-mintlify I think it's ok if we have it defined even though it's defined in our backend. Kinda like how it's defined in npm package.json's too. I haven't put a bunch of thought into it... Answering the question "Is there a benefit to defining the information in mint config even if they're using the Mintlify cloud version?

docs/mint.json
```diff
@@ -151,9 +145,16 @@
...
+    "owner": "mintlify",
+    "repo": "mint",
+    "deployBranch": "main",
+    "path": "/",
```
hahnbeelee: I believe our DB uses "" for base path, not "/", but even then our docs don't live in the base path they live in "/docs"

hahnbeelee: Oh wait actually, looking at our docs it looks like we dont precede it with a "/" so it would just be "docs"

hahnbeelee: Thoughts on naming this "directory" instead? I know we use "path" everywhere, but I'm starting to feel like it is kinda generic/not super intuitive?

docs/mint.json
tobias-mintlify: IMO users should only define if they want to override the DB settings

docs/mint.json
tobias-mintlify: good catch! yea we don't need trailing slashes

client/src/types/config.ts
```diff
@@ -88,6 +88,16 @@ export type Config = {
     icon?: string;
     color?: string;
   };
+  repo?: {
+    github?: {
+      owner: string;
+      repo: string;
+      deployBranch: string;
+      path: string;
+      isPrivate: boolean;
```
hahnbeelee, line 97: You're gonna have to make all of these values optional just because the configuration is user-defined and there's no guarantee that these fields will always exist.

docs/mint.json
```diff
@@ -15,18 +15,15 @@
       "to": "#2AB673"
     }
   },
-  "topbarLinks": [
-    {
-      "name": "Showcase",
-      "url": "https://mintlify.com/showcase"
-    }
-  ],
+  "topbarLinks": [{
```
hahnbeelee, line 18: Can you change this so the formatting is back to how it was before? With the brackets on a different line?

client/src/ui/Feedback.tsx
```diff
@@ -43,85 +42,94 @@ const FeedbackTooltip = ({ message }: { message: string }) => {
...
-          xmlns="http://www.w3.org/2000/svg"
-          viewBox="0 0 512 512"
-        >
-          <path d="M58.57 323.5L362.7 19.32C387.7-5.678 428.3-5.678 453.3 19.32L492.7 58.75C495.8 61.87 498.5 65.24 500.9 68.79C517.3 93.63 514.6 127.4 492.7 149.3L188.5 453.4C187.2 454.7 185.9 455.1 184.5 457.2C174.9 465.7 163.5 471.1 151.1 475.6L30.77 511C22.35 513.5 13.24 511.2 7.03 504.1C.8198 498.8-1.502 489.7 .976 481.2L36.37 360.9C40.53 346.8 48.16 333.9 58.57 323.5L58.57 323.5zM82.42 374.4L59.44 452.6L137.6 429.6C143.1 427.7 149.8 424.2 154.6 419.5L383 191L320.1 128.1L92.51 357.4C91.92 358 91.35 358.6 90.8 359.3C86.94 363.6 84.07 368.8 82.42 374.4L82.42 374.4z" />
-        </svg>
-        <svg
-          className="h-3.5 w-3.5 hidden group-hover:block fill-slate-500 dark:fill-slate-400 group-hover:fill-slate-700 dark:group-hover:fill-slate-200"
-          xmlns="http://www.w3.org/2000/svg"
-          viewBox="0 0 512 512"
+      {createSuggestHref && (
```
hahnbeelee, line 85: Since we're in the midst of editing this code, can you make the below svgs and components more modular by creating components for them with better names?

client/src/ui/MDXContentController/PageHeader.tsx
```diff
@@ -68,9 +68,7 @@ export function PageHeader({ section, pageMetadata }: PageHeaderProps) {
               </p>
             )}
           </div>
-          {/* TODO: Make UserFeedback work in self-hosting
-          https://github.com/mintlify/mint/issues/118 */}
-          {subdomain && <UserFeedback />}
```
hahnbeelee, line 73: Can you link this PR to this issue so that when we merge it it closes this issue?

docs/mint.json
```diff
@@ -156,4 +150,4 @@
       "siteId": "YSVUHCAK"
     }
   }
-}
+}
```
hahnbeelee, line 153: Can you also add the newline again?

client/src/ui/Feedback.tsx
```diff
@@ -43,85 +42,94 @@ const FeedbackTooltip = ({ message }: { message: string }) => {
 export function UserFeedback() {
   const path = useCurrentPath();
   const { mintConfig } = useContext(ConfigContext);
-  const { createSuggestHref, createIssueHref } = useContext(FeedbackContext);
+
+  const [createIssueHref, setCreateIssueHref] = useState<string | null>(null);
+  const [createSuggestHref, setCreateSuggestHref] = useState<string | null>(null);
+
+  useEffect(() => {
+    if (path && mintConfig?.repo?.github && !mintConfig.repo.github.isPrivate) {
```
hahnbeelee, line 50: Along with what I said of making all the properties inside the github property optional I think you're going to need to make this `mintConfig?.repo?.github?.isPrivate` for safety reasons of the property not existing

client/src/ui/Feedback.tsx
```diff
@@ -43,85 +42,94 @@ const FeedbackTooltip = ({ message }: { message: string }) => {
...
+      const { owner, repo, deployBranch, path: deploymentPath } = mintConfig.repo.github;
```
hahnbeelee, line 51: Same here. It's possible this spreading of props will fail if the user forgot to set these properties.

client/src/ui/Feedback.tsx
```diff
@@ -43,85 +42,94 @@ const FeedbackTooltip = ({ message }: { message: string }) => {
...
+      if (deploymentPath == null) {
+        urlPath = 'docs/';
+      } else if (deploymentPath === '') {
+        urlPath = '';
+      } else {
+        urlPath = `${deploymentPath}/`;
+      }
+
+      setCreateSuggestHref(
+        `https://github.com/${owner}/${repo}/edit/${deployBranch}/${urlPath}${removeFirstSlash(
```
hahnbeelee, line 65: Instead of appending a slash to urlPath and removing a slash from path, why don't we error handle for `urlPath` so that we can guarantee that all leading and following slashes are removed and do
`${urlPath}/${path}.mdx`?

hahnbeelee (changes requested):

client/src/types/config.ts
hahnbeelee: And the default behavior would be to assume it's public

client/src/ui/Feedback.tsx
hahnbeelee: Since `urlPath` is user-defined I think the easiest way to normalize slashes is to use the [path module](https://nodejs.org/api/path.html#pathnormalizepath)

client/src/ui/Feedback.tsx
```diff
@@ -43,85 +42,94 @@ const FeedbackTooltip = ({ message }: { message: string }) => {
...
+      // TO DO: We should add the / at the end when not set, but gracefully handle when users
+      // add it themselves.
+      let urlPath;
+      if (deploymentPath == null) {
+        urlPath = 'docs/';
+      } else if (deploymentPath === '') {
+        urlPath = '';
+      } else {
+        urlPath = `${deploymentPath}/`;
+      }
```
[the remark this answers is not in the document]
tobias-mintlify: ^ Could make the default `""` as well

client/src/ui/Feedback.tsx
naz-mintlify: Correct me if I'm wrong, but this path module is on Nodejs and not in react. 

useCurrentPath() returns trailing slash in front always.

tobias-mintlify: ^ Nawaz is right, the path module won't work here.

client/src/types/config.ts
```diff
@@ -88,6 +88,16 @@ export type Config = {
...
+  repo?: {
+    github?: {
+      owner: string;
+      repo: string;
+      deployBranch: string;
+      contentDirectory: string;
+      isPrivate?: boolean;
+    };
+  };
+
```
tobias-mintlify, line 100: please remove extra blank line

client/src/ui/Feedback.tsx
hahnbeelee: Oh shoot you're right. I see the edge cases now.

hahnbeelee (approved):

hahnbeelee:
Closes #118

PR #312, opened 2023-02-02
Nawaz/min 1373 allow search to work on self hosted

naz-mintlify:
## Summary
-  Add Algolia env vars 

## Test Plan
- If Algolia env vars are added, then searchbox should show and work. Otherwise it should not.


hahnbeelee:
In order for this to work, we need to replicate the prebuild step, more specifically:
> Replicate the logic to index the results (turn pages into sections that are then added to Algolia) in the prebuild step on the client side

I will close this PR for now and we can revisit it after we implement the client-side prebuild step.

PR #315, opened 2023-02-04
Use slugToTitle when missing titles

hanywang2:
## Summary

We use slugToTitle to generate titles for missing pages and navigation but not usePrevNext. This PR fixes that

Before
CleanShot 2023-02-03 at 16 38 20@2x

After
CleanShot 2023-02-03 at 16 37 18@2x

## Test Plan

Go to a page without a `title` (most open an openapi page) like [telecomsxchange.localhost:3000](http://telecomsxchange.localhost:3000) and see that the prev next titles are available


tobias-mintlify (changes requested):
`getAllMetaTags` has `pageMeta.title || pageMeta.sidebarTitle || slugToTitle(pageMeta.href || '')` could we take that line and break it into its own "default title" function then use it in both places?

tobias-mintlify (approved):

PR #319, opened 2023-02-04
Unified tsconfig

benjamin-shen:
## Summary
Switch to a unified [tsconfig.json](https://github.com/mintlify/config/blob/main/packages/ts-config/tsconfig.json). There are no direct changes to the configs in this project, but the unified tsconfig.json may be more opinionated in certain areas.

## Test Plan
Every project should build.

## Notes
- We should standardize our `outDir`. I set the default to `dist`.
- The CLI package is not on the unified tsconfig.json because it's too strict.


client/tsconfig.json
```diff
@@ -1,24 +1,10 @@
...
-    "isolatedModules": true,
-    "jsx": "preserve",
-    "incremental": true,
     "baseUrl": ".",
     "paths": {
       "@/*": ["src/*"]
     }
   },
-  "include": ["**/*.js", "**/*.ts", "**/*.tsx"],
-  "exclude": ["node_modules"]
```
benjamin-shen, line 23: no change

packages/validation/tsconfig.json
```diff
@@ -1,14 +1,7 @@
...
   "compilerOptions": {
-    "strict": true,
-    "target": "es5",
-    "moduleResolution": "node",
-    "outDir": "dist",
-    "declaration": true,
-    "allowJs": true,
-    "esModuleInterop": true,
-    "allowSyntheticDefaultImports": true,
-    "baseUrl": "."
```
benjamin-shen, line 11: no change

client/tsconfig.json
```diff
@@ -1,24 +1,10 @@
...
-    "strict": true,
-    "forceConsistentCasingInFileNames": true,
-    "noEmit": true,
-    "esModuleInterop": true,
-    "module": "esnext",
-    "moduleResolution": "node",
-    "resolveJsonModule": true,
-    "isolatedModules": true,
-    "jsx": "preserve",
-    "incremental": true,
```
benjamin-shen, line 16: no change

packages/mintlify/package.json
```diff
@@ -47,6 +47,8 @@
     "yargs": "^17.6.0"
   },
   "devDependencies": {
+    "@mintlify/ts-config": "1.0.4",
+    "@tsconfig/next": "1.x",
```
hahnbeelee, line 51: do we need the next tsconfig for the cli?

hahnbeelee: This question applies for all the packages

hahnbeelee:
Do we want to make the unified tsconfig less frontend/next.js-oriented?

packages/mintlify/package.json
benjamin-shen: @hahnbeelee it's just a base configuration that included a lot of things that I liked. There were next-specific ones (such as module/moduleResolution and lib), but also non-next-specific ones. I agree that our unified tsconfig should be less opinionated, so I changed it to extend `@tsconfig/recommended` instead and included some of the options that I liked.

Side note: The downside of this unified setup is that you can't extend multiple tsconfigs (no diamond problem here 🤡) so we can't do something like `"extends": ["@mintlify/ts-config", "@tsconfig/next"]`. They are adding this feature in TS 5.0, though.

hahnbeelee: Gotcha! Makes sense.

O interesting. RIP. extending multiple tsconfigs seems awesome. Glad to know they are working on supporting it.

packages/mintlify/tsconfig.json
```diff
@@ -1,4 +1,5 @@
 {
+  // "extends": "@mintlify/ts-config",
```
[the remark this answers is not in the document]
hahnbeelee: Were there a lot? :bufo-sus:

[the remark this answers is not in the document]
benjamin-shen: Yes :bufo-sad-vibe:

packages/validation/tsconfig.json
```diff
@@ -1,14 +1,9 @@
...
-    "strict": true,
-    "target": "es5",
     "moduleResolution": "node",
-    "outDir": "dist",
     "declaration": true,
-    "allowJs": true,
-    "esModuleInterop": true,
-    "allowSyntheticDefaultImports": true,
-    "baseUrl": "."
+    "noEmit": false
```
benjamin-shen, line 6: Wait. how'd this get here :bufo-sus:

hahnbeelee (approved):
client
- [x] `yarn dev`
- [x] `mintlify.localhost:3000`
- [x] `yarn build`

mintlify
- [x] `yarn local`
- [x]  `mintlify dev`

validation
- [x] `yarn build`
- [x] `yarn test`

PR #320, opened 2023-02-06
Add More Event Logs for Analytics 1.0

hanywang2:
## Summary

Add logging for the following events when a user 
- Clicks on the CTA
- Clicks on the navigation item
- Copies a code snippet (from the codeblock)
- Makes an API call
- Scrolls to the bottom of the page

## Test Plan

Make sure to help test this one thoroughly. To do so, add the following code snippets to L135-136 of `AnalyticsMediator.ts`

```javascript
console.log(eventName);
console.log(eventProperties);
```

CleanShot 2023-02-05 at 16 42 47@2x

Then click around for the following events and you should see the above events get triggered properly.

**Note**: Make sure to double check that things aren't getting double logged

### DIFF
--- a/client/src/ui/Header.tsx
+++ b/client/src/ui/Header.tsx
@@ -126,7 +126,7 @@ function GitHubCta({ button }: { button: TopbarCta }) {
         href={button.url}
         target="_blank"
         rel="noreferrer"
-        onClick={() => trackCtaClick({ url: 'button.url', type: 'github' })}
+        onClick={() => trackCtaClick({ url: button.url, type: 'github' })}
       >
         <div className="group flex items-center space-x-3">
           <svg
--- a/client/src/ui/Header.tsx
+++ b/client/src/ui/Header.tsx
@@ -8,7 +8,9 @@ import Router from 'next/router';
 import { useContext } from 'react';
 import { useEffect, useState } from 'react';
 
+import AnalyticsContext from '@/analytics/AnalyticsContext';
 import { ConfigContext } from '@/context/ConfigContext';
+import { Event } from '@/enums/events';
 import { Logo } from '@/ui/Logo';
 import { SearchButton } from '@/ui/search/Search';
 import getLogoHref from '@/utils/getLogoHref';
@@ -17,8 +19,6 @@ import { TopbarCta } from '../types/config';
 import Icon from './Icon';
 import { ThemeToggle } from './ThemeToggle';
 import { VersionSelect } from './VersionSelect';
-import AnalyticsContext from '@/analytics/AnalyticsContext';
-import { Event } from '@/enums/events';
 
 export function NavPopover({
   display = 'md:hidden',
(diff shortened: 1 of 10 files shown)


hahnbeelee:
Thank you for the test plan uwu :3

hahnbeelee (changes requested):
I scrolled to the bottom of the page and I didn't get a console log on http://mintlify.localhost:3000/quickstart

image

hahnbeelee (changes requested):
- [x] Clicks on the CTA
- [x] Header nav item click
- [ ] Clicks on the navigation item - nothing is console logging for me
- [x] Copies a code snippet (from the codeblock)
- [x] Makes an API call
- [x] Accordion open + close
- [ ] Scrolls to the bottom of the page
^ Didn't work on http://mintlify.localhost:3000/quickstart
Logged scroll_to_bottom twice on http://mintlify.localhost:3000/components/accordion
image

hahnbeelee:
In the future I think we should architect this in a way so that we can define the inputs for the types for the different event handlers.

hahnbeelee:
Some ideas for events to track:
- Clicking on an anchor
- Might be worth to figure out how to take the title of code blocks as inputs? Not sure how difficult implementing this would be.

client/src/ui/Header.tsx
```diff
@@ -180,6 +192,7 @@ function TopBarCtaButton({ button }: { button: TopbarCta }) {
           href={button.url ?? '/'}
           target="_blank"
           className="group px-4 py-1.5 relative inline-flex items-center rounded-full shadow-sm text-sm font-medium"
+          onClick={() => trackCtaClick({ name: button.name, url: button.url })}
```
tobias-mintlify, line 195: Should we pass a `type` event property here? Maybe `type="button"`?

client/src/ui/MDXContentController/MDXContentController.tsx
```diff
@@ -75,6 +77,31 @@ export function MDXContentController({
...
+  // Log event when scroll to bottom
+  useEffect(() => {
+    if (checkIfAtBottomOfPage()) {
+      trackScrollToBottom({ path: Router.asPath });
+      return;
+    }
+
+    const onScroll = () => {
+      if (checkIfAtBottomOfPage()) {
+        trackScrollToBottom({ path: Router.asPath });
```
tobias-mintlify, line 97: `asPath` will include other things in the URL like the header ID when permalinked: `/my-page#header-id`. You may want to use `useCurrentPath()` and pass that to useEffect. Passing `useCurrentPath` as a dependency might also fix a potential bug I mention in the next comment.

client/src/ui/MDXContentController/MDXContentController.tsx
```diff
@@ -75,6 +77,31 @@ export function MDXContentController({
...
+        window.removeEventListener('scroll', onScroll);
+      }
+    };
+    window.addEventListener('scroll', onScroll);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
```
tobias-mintlify, line 103: Please test this carefully. I'm not sure the event will fire multiple times if you scroll to the bottom of multiple pages. The listener might be removed after the first time and never placed again.

client/src/ui/search/Search.tsx
```diff
@@ -305,7 +310,13 @@ export function SearchProvider({
       .toLowerCase()
       .replaceAll(' ', '-')
       .replaceAll(/[^a-zA-Z0-9-_#]/g, '');
-    router.push(`/${hit.slug}${sectionSlug}`);
+    router.push(`${path}${sectionSlug}`);
```
tobias-mintlify, line 313: Please avoid using `path` as a variable name. We use that in too many places. It's okay to merge this for now though.

PR #322, opened 2023-02-06
normalize z-index classes and fix stacking context

NicoZweifel:
## Summary
Fixes the remaining z-index Issues (https://github.com/mintlify/components/issues/42) by ordering the StackingContext and seperating responsibilities of z-index classes like suggested in https://github.com/mintlify/components/pull/44#issuecomment-1407466150.

Information about how the StackingContext and z-index behaves can be found [here](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Positioning/Understanding_z_index/The_stacking_context).
We don't have to remove `sticky`, but if creating a new StackingContext with `sticky`, it needs to have at least the same z-index if it is intended to display elements (tooltips) above, for example, the navbar.

I split the responsibilities of the zIndex Classes by adding this, which should help in the future with these kind of issues. 
```ts
export const zIndex = {
  Overlay: 'z-50',
  Popup: 'z-40',
  Navbar: 'z-30',
  SideBar: 'z-20',
  Control: 'z-10',
};
```

## Test Plan
Tested locally with the mint docs.


client/src/layouts/ZIndex.ts
```diff
@@ -0,0 +1,7 @@
+export const zIndex = {
```
tobias-mintlify, line 1: Tailwind may not be detecting these classes now that you moved them out of the React code. We usually need to add classes to `safelist` at the bottom of the Tailwind config when we programatically generate the class.

tobias-mintlify (changes requested):
_Love_ the idea of making `ZIndex.ts` it'll make it much easier to maintain everything.

There is a bug where the nav bar is showing underneath the content though:

navbar

I checked the classes with inspect element and "z-30" was set but didn't have a class definition. Tailwind is not detecting the class. You should try changing the `ZIndex.ts` to a `.tsx` file so Tailwind will scan it. If that still doesn't work, add the classes to the allowlist in the Tailwind config.

Lastly, there were two or three other places with a z-10 and z-50 index we should switch to the ZIndex object. One of them was in the search bar code.

NicoZweifel:
Very good points, I'll improve the PR with the things you mentioned.

NicoZweifel:
Changing to .tsx did the trick.
I guess this also explains why it was working originally when I tested locally, probably the css was still cached in the browser.

I also improved the type inference and eslint warnings in MDXContentController a bit but the params and remaining types should probably be handled in a different issue.

Usually these warnings are fixable with [these](https://www.typescriptlang.org/docs/handbook/utility-types.html) and/or with generics.

client/src/layouts/ZIndex.ts
NicoZweifel: I changed the remaining z-index classes after searching in all files, they should all be normalized now.
Resolving for now, If adding it to the config is prefered over using a .tsx file, I could change the PR.

tobias-mintlify (approved):
Thank you this looks great! I tested and all code blocks with names and code groups are working correctly. I found a bug where code blocks without a file name have the same issue, but we likely just need to set the z-index in mintlify/components to fix this.

You can see code blocks without a file name here: http://localhost:3000/writing-content/code

no file name

NicoZweifel:
I checked and it seems to be because of  `absolute` and `z-10`  on the parent, creating a new StackingContext:
image

PR #329, opened 2023-02-07
Revert sidebar title change

tobias-mintlify:
The getTitle function gets the full title. The sidebar is the one time `sidebarTitle` is prioritized.


hahnbeelee (changes requested):
Either delete getTitle or modify it (most likely modify it because the same logic should be used everywhere)

hahnbeelee (approved):
This is an edge case where the sidebar uses the sidebar title.
We should discuss if we want to normalize the side bar title for usePrevNext

PR #333, opened 2023-02-07
Create sidebarTitle helper function

tobias-mintlify:
Also starts using sidebar titles for footer navigation.


hahnbeelee (approved):
slay

PR #339, opened 2023-02-08
Address trivial errs for cli

hahnbeelee:
## Summary
I didn't try to address everything because that would be very ambitious.
Before:
image

After:
image

## Test Plan
* Look over code and make sure the logic is sound.
* Test basic CLI functionality


packages/mintlify/src/downloadImage.ts
```diff
@@ -62,7 +62,7 @@ export function isValidImageSrc(
   const imageHref = removeMetadataFromImageSrc(src);
   const ext = getFileExtension(imageHref);
 
-  if (!SUPPORTED_MEDIA_EXTENSIONS.includes(ext)) {
+  if (ext && !SUPPORTED_MEDIA_EXTENSIONS.includes(ext)) {
```
tobias-mintlify, line 65: Maybe we can remove the `skipValidateImageExtension` check now that you added this?

packages/mintlify/src/scraping/detectFramework.ts
```diff
@@ -13,14 +13,15 @@ export function detectFramework(html) {
 
   if (
     docusaurusMeta.length > 0 &&
-    docusaurusMeta.attr("content").includes("Docusaurus")
+    docusaurusMeta.attr("content") &&
+    docusaurusMeta.attr("content") === "Docusaurus"
```
tobias-mintlify, line 17: All these checks have to be `.includes`, `===` is incorrect. The full string is of the form `Docusaurus v3.0.0-alpha.0`.

packages/mintlify/src/scraping/site-scrapers/scrapeDocusaurusSection.ts
```diff
@@ -11,13 +11,15 @@ export async function scrapeDocusaurusSection(
...
-  version: string
+  version?: string
 ): Promise<MintNavigation[]> {
   const $ = cheerio.load(html);
 
   // Download the logo
   const logoSrc = $(".navbar__logo img").attr("src");
-  downloadLogoImage(logoSrc, imageBaseDir, origin, overwrite);
+  if (logoSrc) {
+    downloadLogoImage(logoSrc, imageBaseDir, origin, overwrite);
```
tobias-mintlify, line 21: `downloadLogoImage` already exits without downloading if logoSrc is missing. I'd prefer to keep it the way we had it to avoid duplicate if-statements in every scraper implementation.

packages/mintlify/src/util.ts
```diff
@@ -108,7 +108,7 @@ export const createPage = (
         flag: "wx",
       });
       console.log("✏️ - " + writePath);
-    } catch (e) {
+    } catch (e: any) {
```
tobias-mintlify, line 111: You don't need to set a type when catching things in Typescript. In fact, the only types it would set let you set are `any` and `unknown`. Keeping the default seems better than disabling all type checking

tobias-mintlify (changes requested):

packages/mintlify/src/downloadImage.ts
hahnbeelee: What was the reasoning for adding it in the first place? I see [the PR](https://github.com/mintlify/mint/pull/284) Does intercom not have extensions?

tobias-mintlify: Yeah Intercom doesn't have extensions so the `SUPPORTED_MEDIA_EXTENSIONS.includes` check fails. Your `ext &&` addition to this if-statement addresses the issue in a simpler way.

packages/mintlify/src/scraping/site-scrapers/scrapeDocusaurusSection.ts
hahnbeelee: The beauty of typescript is that we don't have to do `string | undefined` as inputs for functions. Sometimes it's ok to add if guards if it means that the logic is more clear to follow.

tobias-mintlify: In that case, you need to update the if-statements to do the full check: `if (!isValidImageSrc(imageSrc))`

packages/mintlify/src/scraping/downloadAllImages.ts
```diff
@@ -33,7 +33,7 @@ export default async function downloadAllImages(
   // Wait to all images to download before continuing
   const origToNewArray = await Promise.all(
     imageSrcs.map(async (imageSrc: string) => {
-      if (!isValidImageSrc(imageSrc)) {
+      if (imageSrc || !isValidImageSrc(imageSrc)) {
```
tobias-mintlify, line 36: This is going to exit whenever we have a valid image source. Are you missing a `!`?

tobias-mintlify (changes requested):

packages/mintlify/src/scraping/downloadAllImages.ts
hahnbeelee: omg good catch

packages/mintlify/src/scraping/site-scrapers/scrapeDocusaurusSection.ts
hahnbeelee: I just changed it to `string | undefined` for now cuz I think it's really subjective. What looks repetitive to you might look cleaner to me etc. but it's really just a small stylistic difference. If I made it so it just took `string` and not `string | undefined`, then `if (isValidImageSrc(imageSrc))` would yell because `imageSrc` would be type `string | undefined`. I think what I have now is good because it's the middle ground of what you were suggesting and what I had originally. There's like 3 layers of this prop being passed down to `isValidImageSrc` and in the second layer we check if it's defined or not which allows us to reduce the amount of repeated code. :3

tobias-mintlify (approved):

packages/mintlify/src/downloadImage.ts
```diff
@@ -43,10 +43,6 @@ async function writeImageToFile(
 }
 
 export function isValidImageSrc(src: string) {
-  if (!src) {
-    return false;
```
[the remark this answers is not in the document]
hahnbeelee: u right. I wasn't thinking.
Since the type is `string`, `undefined` will never be passed to this function so we no longer have to check for `undefined`. However I didn't think about the falsy value  `""`.
For functions that take in arrays we actually shouldn't be checking for `Array.isArray` because if you type it as an array you're allowed to assume it's an array. A check for `arr.length > 0` is good though.

PR #340, opened 2023-02-09
Move canonical tags into separate component

tobias-mintlify:
## Summary

Canonical tag URLs were wrong on first server-side render, causing search engines to see wrong canonical URLs.

`useCurrentPath()` depends on ConfigContext to get the subdomain. The path was wrong before because subdomain is undefined when we are calling that hook from outside the ConfigContext Provider. We moved everything into a new component so we can call useCurrentPath correctly. The new code uses <Head> instead of being placed into the existing <Head> tag because Next Router threw an error when I put the component in <Head>

## Test Plan

Tested with `yarn dev` and `yarn dev-watch`

Console logging is the only reliable way to test because the URLs are correct after the first render.


client/src/ui/CanonicalTags.tsx
```diff
@@ -0,0 +1,22 @@
...
+import { useState, useEffect } from 'react';
+
+import { useCurrentPath } from '@/hooks/useCurrentPath';
+
+export const CanonicalTags = () => {
+  const [origin, setOrigin] = useState('');
+
+  useEffect(() => {
+    setOrigin(window.location.origin);
+  }, []);
```
hahnbeelee, line 11: Looks like this code is used in 2 places. I wonder if we can create some reusable fetcher or context for this

client/src/layouts/SupremePageLayout.tsx
```diff
@@ -14,12 +14,11 @@ import { useAnalytics } from '@/analytics/useAnalytics';
 import components from '@/components';
 import { ConfigContext } from '@/context/ConfigContext';
 import { VersionContextController } from '@/context/VersionContext';
-import { useColors } from '@/hooks/useColors';
```
hahnbeelee, line 17: What the did eslint not catch this unused import?

client/src/ui/CanonicalTags.tsx
tobias-mintlify: I'm not sure if we'll keep using it in SupremePageLayout. We can refactor it in the near future if we keep using `origin`.

client/src/layouts/SupremePageLayout.tsx
tobias-mintlify: It's a warning, we may have been ignoring it this whole time haha

client/src/ui/CanonicalTags.tsx
hahnbeelee: Can you make a linear ticket for it?

client/src/layouts/SupremePageLayout.tsx
hahnbeelee: https://github.com/mintlify/mint/blob/main/client/.eslintrc.json#L7

hahnbeelee (approved):

PR #348, opened 2023-02-10
Support allOf for OpenAPI specs

hanywang2:
## Summary

This PR enables the setup for `allOf` syntax in OpenAPI, which combines several objects types used by a small handful of customers.

## Test Plan

- Go to `kombo.localhost:3000` and `Create Candidate`. You should be able to see the body params filled in and not missing as on prod.

CleanShot 2023-02-10 at 00 22 32@2x


client/src/utils/openapi.ts
```diff
@@ -44,3 +44,37 @@ export const getProperties = (properties: any[]): any[] => {
...
+
+export const getTypeName = (type: string[] | string) => {
+  return Array.isArray(type) ? type.join(' | ') : type;
+};
+
+export const combineAllOfIntoObject = (allOf: any[]) => {
+  let combinedProperties = {};
+  let combinedRequired: string[] = [];
+  let combinedExample = {};
+  allOf.map((item) => {
```
tobias-mintlify, line 60: This should be a `forEach` instead of `map` because you are not assigning the result to anything.

hanywang2: Valid

tobias-mintlify (approved):

PR #342, opened 2023-02-09
feat: Added validation for iconType

Lolly1150:
Added validation for iconType

PR #349, opened 2023-02-10
feat: added iconType to AnchorLink components

Lolly1150:
## Summary

I had added iconType props to AnchorLink

Request in [this](https://github.com/mintlify/mint/pull/342#pullrequestreview-1291995099)


tobias-mintlify (changes requested):
You need to pass the iconType property where `StyledAnchorLink` is called. The goal is to pass the property from mintConfig to this component so you can set the icon type by updating `mint.json`

client/src/layouts/NavSidebar/Nav.tsx
```diff
@@ -55,6 +55,7 @@ function TopLevelNav({ mobile }: { mobile: boolean }) {
         className="mb-4"
         color={colors.anchors[0]}
         icon={config?.topAnchor?.icon || 'book-open'}
+        iconType={config?.anchors?.[4]?.iconType || 'duotone'}
```
tobias-mintlify, line 58: This should be using `topAnchor` not `.[4]`

The the line above it.

You will likely need to update the type definition and the validation package if we missed this on the last PR.

client/src/ui/Icon.tsx
```diff
@@ -4,7 +4,7 @@ import { IconType } from '@/types/config';
 
 type IconProps = {
   icon: string;
-  iconType?: IconType;
+  iconType?: IconType | string;
```
tobias-mintlify, line 7: We don't want `| string` when there's only a few options. Why did you have to add this type?

tobias-mintlify (changes requested):

packages/validation/src/schemas/config.ts
```diff
@@ -216,23 +216,26 @@ export const configSchema = z.object({
...
-      .enum(["brands", "duotone", "light", "sharp-solid", "solid", "thin"], {
-        errorMap: () => {
-          return {
-            message:
-              "topAnchor.iconType must be one of the following strings: brands, duotone, light, sharp-solid, solid, thin",
-          };
-        },
-      })
-      .optional(),
+        iconType: z
```
tobias-mintlify, line 219: this in indented too deep

client/src/types/config.ts
```diff
@@ -97,6 +97,7 @@ export type Config = {
     name: string;
     icon?: string;
     color?: string;
+    iconType: IconType;
```
tobias-mintlify, line 100: can you move this up one line? So it looks like:

```
icon?: string;
iconType: IconType;
color?: string;
```

client/src/ui/AnchorLink.tsx
```diff
@@ -79,7 +82,7 @@ export function StyledAnchorLink({
     ) : (
       <Icon
         icon={icon.toLowerCase()}
-        iconType="duotone"
+        iconType={iconType ?? 'duotone'}
```
tobias-mintlify, line 85: Can you use || instead of ?? here?

packages/validation/src/schemas/config.ts
```diff
@@ -216,13 +216,26 @@ export const configSchema = z.object({
             invalid_type_error: "topAnchor.icon must be a string",
           })
           .optional(),
+        iconType: z
```
tobias-mintlify, line 219: Please reduce the indenting level by one.

tobias-mintlify (changes requested):

tobias-mintlify (approved):

PR #352, opened 2023-02-10
Set base path in canonical URLs

tobias-mintlify:
## Summary

We were excluding the base path from the canonical URLs. Eg: `site.com/page` instead of `site.com/docs/page`

## Test Plan

Tested with `yarn dev`


hahnbeelee (approved):

hahnbeelee (changes requested):
wait lol let me actually test... :bufo-sus:

hahnbeelee (approved):
Set basePath to "/docs" in my .env
Inspect element:
image
Source code:
image

hahnbeelee:
Looks good when there is no base path too:
image
image

PR #354, opened 2023-02-11
Check for basePath+no basePath for canonical tags

hahnbeelee:
## Summary
Problem:
image
When you "View Source" it was showing the incorrect path

I noticed when console logging the router the value of the path was different during `yarn build` and `yarn start/dev`. The basePath wasn't there during build time. This PR makes it so we account for both during build and run time.

## Test Plan
- [x] Run `yarn build && yarn start` with env var `BASE_PATH="/docs"
  - [x] Check "View Source"
  - [x] Check "Inspect element"
- [x] Run `yarn build && yarn start` with env var ""
  - [x] Check "View Source"
  - [x] Check "Inspect element"


client/src/ui/CanonicalTags.tsx
```diff
@@ -1,20 +1,27 @@
...
-
-  // useCurrentPath needs to be nested inside ConfigContext so we can't calculate the
-  // canonical url in SupremePageLayout.
-  const currentUrl = origin + router.basePath + useCurrentPath();
+  let currentPath = router.asPath.split('#')[0];
+  if (router.asPath.startsWith(withOutBasePathRemoval)) {
+    currentPath = router.asPath.substring(withOutBasePathRemoval.length).split('#')[0];
+  }
+  if (router.asPath.startsWith(withBasePathRemoval)) {
+    currentPath = router.asPath.substring(withBasePathRemoval.length).split('#')[0];
```
tobias-mintlify, line 22: If useCurrentPath is broken we should fix it there instead of adding code to CanonicalTags.

tobias-mintlify (changes requested):

hahnbeelee:
Changed code:
- [x] Run `yarn build && yarn start` with env var `BASE_PATH="/docs"
  - [x] Check "View Source"
image
  - [x] Check "Inspect element"
image

- [x] Run `yarn build && yarn start` with env var ""
  - [x] Check "View Source"
image

  - [x] Check "Inspect element"
image

hanywang2 (approved):

PR #357, opened 2023-02-13
Lint staged git hook

NicoZweifel:
## Summary
Adds `pre-commit` git hook to automatically format and lint staged files.
Will not allow new lint warnings to be commited unless when using `--no-verify`
This will help a lot with maintaining, especially with many contributors, and with reducing the amount of eslint warnings.

It might be a good idea to have reusable, shared eslint configs (typescript -> react -> next) and use git hooks on the other projects, like libraries, component libraries etc.

In this repository it might even make sense to convert it to a monorepo for the purpose of git hooks, testing and maintenance.

## Test Plan
Tested locally, tried to commit a new eslint warning and got usable error output (with navigation link) in the terminal and in github desktop:

image
image


client/package.json
```diff
@@ -128,6 +130,7 @@
     "eslint-plugin-unused-imports": "^2.0.0",
     "gray-matter": "^4.0.3",
     "hast-util-to-string": "^2.0.0",
+    "husky": "^8.0.3",
```
tobias-mintlify, line 133: These should be in devDependencies

NicoZweifel: I think they are? 😋

tobias-mintlify: Ah sorry! I meant to put the comment under `lint-staged`

NicoZweifel: Oh okay 👍 moved

client/package.json
```diff
@@ -137,5 +140,14 @@
     "ts-node": "^10.9.1",
     "typescript": "^4.7.4",
     "unist-util-visit": "^2.0.3"
+  },
+  "lint-staged": {
```
benjamin-shen, line 144: Should we add `"git add"` as an additional command? Otherwise, I think we don't need to wrap the commands in brackets.

client/package.json
```diff
@@ -137,5 +140,14 @@
...
+    "*.+(ts|tsx)": [
+      "eslint --fix --max-warnings 0"
```
benjamin-shen, line 146: Can we use `next lint` instead of `eslint`? Also, is there a reason you omitted `--cache`?

I think we might need to remove the max-warnings, since we still have a bunch of warnings in this repo.

client/package.json
```diff
@@ -137,5 +140,14 @@
...
+    ],
+    "*.{js,jsx,ts,tsx,json,css,md,mdx}": [
```
benjamin-shen, line 148: Same comment as in package.json

benjamin-shen (commented):
Thank you so much @NicoZweifel! This is awesome stuff.

> It might be a good idea to have reusable, shared eslint configs (typescript -> react -> next) and use git hooks on the other projects, like libraries, component libraries etc.

Completely agree, we'll try to get something like that set up soon.

> In this repository it might even make sense to convert it to a monorepo for the purpose of git hooks, testing and maintenance.

This also makes a lot of sense, we've actually been discussing it quite a bit. I appreciate your workarounds in this PR.

client/package.json
NicoZweifel: I tried to do that but if there is no changes in pages directory, it will fail with `no pages directory found` 😞

client/package.json
```diff
@@ -7,11 +7,12 @@
     "build": "next build",
     "start": "next start",
     "export": "next build && next export",
-    "format": "prettier \"**/*.{css,js,ts,tsx,md,mdx}\" --write",
-    "format:check": "prettier \"**/*.{css,js,ts,tsx,md,mdx}\" --check",
+    "format": "prettier \"**/*.{js,jsx,ts,tsx,json,css,md,mdx}\" --write",
+    "format:check": "prettier \"**/*.{js,jsx,ts,tsx,json,css,md,mdx}\" --check",
```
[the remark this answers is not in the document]
NicoZweifel: No reason, alphabetical is fine, I was using a mix of the one in the lint-staged config and the one that exists already.

client/package.json
NicoZweifel: Could you clarify? Do you mean adding another git hook?

client/package.json
NicoZweifel: The max-warnings only applies to new staged files, but if refactoring older code it can become a problem.
Solutions are:
- use `--no-verify` when committing i.e. `git commit --no-verify`
- disable eslint warning for line/file
- fix the warning 😄

NicoZweifel: Regarding cache, we can add or remove any params, it is just a default configuration for our kind of setup.

client/package.json
NicoZweifel: Let's track it in the other comment.

client/package.json
```diff
@@ -137,5 +140,14 @@
...
     "unist-util-visit": "^2.0.3"
+  },
+  "lint-staged": {
+    "*.+(ts|tsx)": [
+      "eslint --fix --max-warnings 0"
+    ],
+    "*.{js,jsx,ts,tsx,json,css,md,mdx}": [
+      "prettier --write"
+    ],
+    "*.js": "eslint --cache --fix"
```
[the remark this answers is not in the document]
NicoZweifel: Let's track it in the other comment.

client/package.json
benjamin-shen: Sure! I meant just adding it to the list of commands. Example: https://github.com/box/box-ui-elements/blob/a88a96cfeaf44c6330284d3ec39072491075e2e3/package.json#L105-L114

I'm honestly fine either way, but if we don't go with it then we can change
```
    "*.+(ts|tsx)": [
      "eslint --fix --max-warnings 0"
    ],
    "*.{js,jsx,ts,tsx,json,css,md,mdx}": [
      "prettier --write"
    ],
    "*.js": "eslint --cache --fix"
```
to
```
    "*.+(ts|tsx)": "eslint --fix --max-warnings 0",
    "*.{js,jsx,ts,tsx,json,css,md,mdx}": "prettier --write",
    "*.js": "eslint --cache --fix"
```

NicoZweifel: Ah yea we could add additional or more seperate commands like those in the example for sure. 
I just tried to keep it as similar to the prettier/lint config used already as possible for now.

There are a bunch of other git hooks that could also be interesting, also some server side ones and github actions which can check the linting results in PRs.

client/package.json
benjamin-shen: Ah, I see.

> I tried to do that but if there is no changes in pages directory, it will fail with no pages directory found 😞
> Regarding cache, we can add or remove any params, it is just a default configuration for our kind of setup.

Let's include the cache flag, and change the cache location to the one used by `next lint` ([reference](https://github.com/vercel/next.js/issues/33096#issuecomment-1007920000)):
```
eslint --cache --cache-location .next/cache/eslint/ --fix
```

> The max-warnings only applies to new staged files, but if refactoring older code it can become a problem.

Can we actually omit it? I think it might discourage contributors from touching the older code. I also don't want the pre-commit hook to be stricter than the CI check.

NicoZweifel: - Okay, I'll update the PR 👍 

- I think for this case there definitely should be a paragraph in contributing.mdx.
It should only happen though if refactoring old code and not resolving it by any of the methods mentioned above, otherwise only new added code is linted on `pre-commit`.

We could also add some eslint rules disables in lines/files that have a lot of warnings to prevent this from happening.

client/package.json
benjamin-shen: > Ah yea we could add additional or more seperate commands like those in the example for sure.

Oh, I actually meant something like
```
    "*.+(ts|tsx)": [
      "eslint --fix --max-warnings 0",
      "git add"
    ],
    "*.{js,jsx,ts,tsx,json,css,md,mdx}": [
      "prettier --write",
      "git add"
    ],
    "*.js": [
      "eslint --cache --fix"
    ]
```
but I'm curious to hear your opinion. I could see pros and cons of both.

> There are a bunch of other git hooks that could also be interesting, also some server side ones and github actions which can check the linting results in PRs.

Absolutely, lots to explore here!

client/package.json
benjamin-shen: 💯 

> I think for this case there definitely should be a paragraph in contributing.mdx.

Agreed! This is definitely one of those things that projects do differently and new contributors might find confusing at first.

> We could also add some eslint rules disables in lines/files that have a lot of warnings to prevent this from happening.

The warnings are honestly just tech debt that we want to eventually get around to fixing, so we are intentionally keeping the warnings visible. Soon they will be gone 😁

NicoZweifel: I will add a small paragraph (setup, help) to contributing.mdx and will mention the solutions  for bypassing/resolving.

I was checking, adding some new types and trying to resolve the warnings with some custom types on a branch a few days ago and most of the warnings are a result of having to deal with the unknown types in `openapi.ts`, which use `any` and the subsequent usage of  `any` in the rest of the component tree.

We should be able to resolve them slowly by adding more type definitions all the way down to `openapi.ts` with the help of [utility types](https://www.typescriptlang.org/docs/handbook/utility-types.html) and [generics](https://www.typescriptlang.org/docs/handbook/2/generics.html).

Especially the utility types are great for resolving these warnings.

NicoZweifel: [This](https://github.com/CompSciLauren/awesome-git-hooks) is also a nice resource to get an idea of the possibilities of git hooks.
For example `commit-msg` has been used quite a lot on projects that I was working on to make sure the commit-msg follows a certain format, for example `fix: button color`, `feat: dropdown` or checking for an issue/ticket reference/number.

There is also some [examples ](https://github.com/CompSciLauren/awesome-git-hooks/blob/master/pre-commit-hooks/spell-check-md-files.hook)for spell checking md files and [this ](https://github.com/CompSciLauren/awesome-git-hooks/blob/master/pre-commit-hooks/verify-name-and-email.hook) one for verifying the name/email is also nice and is used quite a lot on both private and public projects.

client/package.json
NicoZweifel: found this: 

```From v10.0.0 onwards any new modifications to originally staged files will be automatically added to the commit. If your task previously contained a git add step, please remove this...```

from [here](https://github.com/okonet/lint-staged#v10).

Seems like the `git add` is just a leftover in some example configs and is not needed any longer.
I didn't need `git add` in the past, are there some cases where it would still be required?

Comment (benjamin-shen) [reply client/package.json:146]: Oh nice! I agree, it's probably just leftover code.

File: client/package.json [hunk 7]
```diff
@@ -139,5 +142,13 @@
...
     "typescript": "^4.7.4",
     "unist-util-visit": "^2.0.3"
+  },
+  "lint-staged": {
+    "*.+(ts|tsx)": [
+      "eslint --cache --cache-location .next/cache/eslint/ --fix --max-warnings 0"
+    ],
+    "*.{css,js,ts,tsx,md,mdx}": [
+      "prettier --write"
+    ]
```
benjamin-shen, line 152: Does this work?

```suggestion
    "*.+(ts|tsx)": "eslint --cache --cache-location .next/cache/eslint/ --fix",
    "*.{css,js,ts,tsx,md,mdx}": "prettier --write"
```

client/package.json
NicoZweifel: If it is fine with you, I'll resolve this convo and we leave the arrays for easy access when adding new commands.

benjamin-shen (approved):

NicoZweifel:
I added some documentation in https://github.com/mintlify/mint/pull/367

client/package.json
NicoZweifel: Yea I just tested and comitted the change, it works as well.
The only reason to leave the arrays would be that it is easier to add a new script if somebody ist not familiar with `lint-staged` but either way is completely fine with me

client/.gitignore
```diff
@@ -47,3 +47,6 @@ yarn-error.log*
 
 # Sentry
 .sentryclirc
+
+# eslint
+.eslintcache
```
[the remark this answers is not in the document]
NicoZweifel: removed

benjamin-shen:
@NicoZweifel  ready to merge?

NicoZweifel:
Yea, ready to go 🚀

PR #359, opened 2023-02-13
Split MdxContentController state, context, utilities and UI 

NicoZweifel:
## Summary
When I first started working on mint I found it a bit hard to read the MDXContentController because of a lot of `any` usage and state, context, utilities and UI mixed in the same file.

This PR moves state, hooks and utility functions into seperate files and makes it possible to extend the existing context or add something like MDXContentController context to seperate the component without having to pass props.

- ~~Could add MDXContentController context and a reducer, split components and then it could look similar to this~~:
```tsx
const [state,dispatch] = useMDXContentController(props);
return (
<MDXContentController value={[state,dispatch]}>
  <MDXContentControllerHeader />
  <MDXContentControllerContent />
  <MDXContentControllerFooter />
  <MDXContentControllerSideBar />
<MDXContentController />
);
```
- Children could also have a hook to access MDXContent context like this:
```tsx
const [state] = useMDXContent();
```

I think this would be much cleaner, help new contributors in getting familiar with the MDXContentController and help overall with improving types, testing possibilities and reducing eslint warnings.

## Test Plan

Draft PR for now, could add extra tests for new isolated functions.


NicoZweifel:
Added some more changes to demonstrate, could also add a reducer to simplify the state in useMDXContentController.

tobias-mintlify:
I'm loving this. 🥇 Let me know if you need any help I'm excited to ship this refactor

NicoZweifel:
I think this is now in it's final stages, I just added a reducer and some more hooks + dispatches with dependencies.
This should reduce the amount of re-renders that are triggered.
I did not add reducer functions for every field but there is a dispatch to update the whole state with a Partial type.

I will test some more and possibly add some tests for the state/hooks.

NicoZweifel:
@tobias-mintlify Would be happy to have another set of eyes to take a look and make some suggestions/improvements.

We could always move more stuff out of `UseMDXContentController` into seperate hooks/files but I think with ~100 lines it's fine.

Other than that I think it is worth it to improve some of the types to remove some of the `any` occurrences, but I think this should probably be done in a seperate PR.

NicoZweifel:
Naming could also be improved probably, for example I was thinking of renaming `useGeneratedRequestExamples` to a more generic `useContentApi` or something similar, since it also updates the `api` and another field. (maybe `useRequestApi`).

tobias-mintlify (commented):
I tested this on a few things and its all functional. 🥳  I'll ask the others to review too when you mark it as ready.

I have a few comments, not related to this PR but things you should be aware of when refactoring:

1. We want apiBaseIndex to be stored while the user navigates between pages. We can keep it like we have right now, but we'll need to load it from localstorage at some point.
2. We want API authentication inputs to be stored between pages. These are slightly more sensitive so localstorage may not be appropriate. It looks to me like this refactor will make it easier to store the authentication values in `state` so we can build this feature.
3. We want to expand Mintlify to power other types of communication like blogs and changelogs. (You've probably seen some WIP blog code). This will likely require us to customize the navigation on the left based on the page mode. This PR looks easier to maintain long-term, just take care we are not locking ourselves in to something that will be hard to customize for other use cases. Having skimmed the PR, I think it's good and any improvements like I'm mentioning can probably be a follow up refactor.
4. Can you write a `contributing.mdx` explaining the dispatch architecture you've added here? If you put it in the `docs` folder it'll show up as a hidden page at `mintlify.com/docs/contributing`. Hidden pages don't show up in the navigation but are available when linked to directly. We'll keep it hidden until we publish the self-hosting docs we're writing.

NicoZweifel:
1. We could add a context + provider to _app.tsx. It is a really convenient way of keeping state between pages in nextjs.
   One thing to keep in mind tho is that next 13 is adapting a new folder structure ([appDir)](https://beta.nextjs.org/docs/routing/pages-and-layouts)  where state between pages is not shared in `app.tsx` anymore but in shared layouts, which would require to move the providers there when adapting the new structure.

2. For Authentication, the same as above applies, although might wanna use a database session or if it is desired to keep authentication between different pages/subdomains, a JWT or cookie might be required.
I am not sure how the authentication works at the moment but [`next-auth`](https://next-auth.js.org/) is super useful when it comes to any form of authentication with next.

3. Yea I saw some of the WIP Blog Code and was wondering about if it is supposed to be SideContent but figured it should be because of the `!isWideSize` check.

4. I will add some comments and documentation tomorrow and mark the PR as ready afterwards 👍

tobias-mintlify:
Regarding `_app.tsx`, we can't add context providers there because multitenant hosting needs to have different values for each "site". Putting context in `SupremePageLayout` should have the intended effect without causing issues.

client/src/ui/MDXContentController/Api.tsx
```diff
@@ -0,0 +1,52 @@
...
+      generatedRequestExamples,
+      pageMetadata,
+      api,
+      openApiPlaygroundProps,
+      paramGroups,
+    },
+    dispatch,
+  ] = useMDXContent();
+
+  const setApiBaseIndex = useCallback(
```
NicoZweifel, line 20: Could also remove the props of  `ApiPlayground` , use 
```tsx 
const [state] = useMDXContent()
``` 
in  `ApiPlayground`  and move these callbacks to `ApiPlayground` .

client/src/ui/MDXContentController/Header.tsx
```diff
@@ -0,0 +1,15 @@
+import { useMDXContent } from '@/hooks/useMDXContent';
+import { BlogHeader, PageHeader } from '@/ui/MDXContentController/PageHeader';
+import { getSectionTitle } from '@/utils/paths/getSectionTitle';
+
+export const Header = () => {
+  const [{ currentPath, mintConfig, pageMetadata, isBlogMode }] = useMDXContent();
+  return isBlogMode ? (
+    <BlogHeader pageMetadata={pageMetadata} />
```
NicoZweifel, line 8: Could also do the same here (remove props, use context in child).

client/src/ui/MDXContentController/SidePanel.tsx
```diff
@@ -0,0 +1,43 @@
...
+          {responseExample}
+          {!responseExample && pageMetadata.openapi && (
+            <OpenApiResponseExample openapi={pageMetadata.openapi} />
+          )}
+        </div>
+      </ContentSideLayout>
+    ) : isBlogMode ? (
+      <BlogContext />
+    ) : (
+      <TableOfContents
```
NicoZweifel, line 36: Could remove the props and use context in children here as well

client/src/ui/MDXContentController/Api.tsx
NicoZweifel: for example like [this](https://github.com/NicoZweifel/mint/blob/remove-props/use-context/client/src/ui/ApiPlayground.tsx):
```tsx
export function ApiPlayground() {
  const { basePath } = useRouter();
  const { mintConfig, openApiFiles } = useContext(ConfigContext);
  const [apiBaseIndex, setApiBaseIndex] = useState(0);

  const [{ openApiPlaygroundProps, pageMetadata, api, paramGroups }] = useMDXContent();

  const contentType = openApiPlaygroundProps.contentType ?? pageMetadata.contentType;

  const onApiBaseIndexChange = useSetApiBaseIndexCallback();
  const onInputDataChange = useSetApiPlaygroundInputsCallback();
...
```

tobias-mintlify (commented):
Tested again and everything looked good. 🚀 I'll wait for some of the others to review before approving since it's a big change.

NicoZweifel:
I added some docs in docs/code/MDXContentController and linked to them from contributing.mdx.
If another structure is preferred, I can change it accordingly.

Would also be eager to hear your opinions about removing props from `SidePanel`, `Header` and `ApiPlayground` and using the context instead.
For example like [here](https://github.com/NicoZweifel/mint/pull/1).

It'd be great to get another review.
I don't feel 100% satisfied with the MDXContentController  docs at the moment (I'd like to use less CodeBlocks/Code for maintenance reasons but still be able to show some concepts.

Feel free to make suggestions.

client/src/context/MDXContentContext.tsx
```diff
@@ -0,0 +1,122 @@
...
+  action
+) => {
+  const type = action.type;
+  switch (type) {
+    case MDXContentActionEnum.SET_STATE:
+    case MDXContentActionEnum.SET_USER_DEFINED_EXAMPLES:
+    case MDXContentActionEnum.SET_TABLE_OF_CONTENTS:
+    case MDXContentActionEnum.SET_CONTENT_WIDTH:
+    case MDXContentActionEnum.SET_PARAM_GROUPS:
+    case MDXContentActionEnum.SET_API_PLAYGROUND:
```
benjamin-shen, line 85: I think `SET_USER_DEFINED_EXAMPLES`...`SET_API_PLAYGROUND` should be stricter on which properties in state are updated, and be in their own case statements.

For example, here is what I would change one of the actions to:
```tsx
case MDXContentActionEnum.SET_USER_DEFINED_EXAMPLES:
  const { requestExample, responseExample } = action.payload;
  return {
    ...state,
    requestExample,
    responseExample,
  };
```

If we don't do this, then we can pass `isApi` (or `isBlogMode`, etc) as an object in the payload so that we can update the state in the same way:
```tsx
dispatch({
  type: MDXContentActionEnum.SET_IS_API,
  payload: { isApi },
});
```
instead of how it's currently done:
```tsx
dispatch({
  type: MDXContentActionEnum.SET_IS_API,
  payload: isApi,
});
```

But then, there is not really a point to distinguishing different actions, and we might as well use `useState` instead of `useReducer`. Of course, I think `useReducer` makes a lot of sense.

Did that make sense? Btw, if this was just a one-to-one mapping of the existing functionality, then I think it can also be cleaned up in another PR. This diff is plenty large 😋

client/src/context/MDXContentContext.tsx
```diff
@@ -0,0 +1,122 @@
...
+import {
+  ApiPlaygroundType,
+  ContentWidthType,
+  MDXContentState,
+  ParamGroupsType,
+  TableOfContentsType,
+} from '@/types/mdxContentController';
+import { OpenApiPlaygroundProps } from '@/ui/MDXContentController/getOpenApiPlaygroundProps';
+
+const initialState = {
```
benjamin-shen, line 13: Can you define a type for this?

client/src/ui/MDXContentController/Footer.tsx
```diff
@@ -50,15 +44,20 @@ const Social = ({ type, url }: SocialProps) => {
...
+  const next = pageMetadata.hideFooterPagination ? null : n;
+  const hasBottomPadding = !isApi;
   return (
     <footer className={clsx('text-sm leading-6', previous || next ? 'mt-12' : 'mt-16')}>
       {(previous || next) && (
         <div className="mb-10 text-slate-700 font-semibold flex items-center dark:text-slate-200">
           {previous && (
             <Link
-              href={previous.href}
+              href={previous?.href as never}
```
benjamin-shen, line 60: Learning question: why `as never`?

benjamin-shen (commented):
Impressive PR 🥇, thank you so much!! This refactor makes a lot of sense to me.

client/src/context/MDXContentContext.tsx
NicoZweifel: Thanks for your input 👍 

Yea, I agree, that makes a lot of sense and I was thinking about these things in similar ways.

>For example, here is what I would change one of the actions to:

I wish there was an easier way to purge the properties in typescript when assigning a type but I don't think there is without using some external frameworks/libraries.

I can implement it with destructuring for `SET_USER_DEFINED_EXAMPLES` and `SET_API_PLAYGROUND`.

>But then, there is not really a point to distinguishing different actions, and we might as well use useState instead of useReducer. Of course, I prefer useReducer.

Yea I think it makes more sense to take advantage of beeing able to change the types for the dispatch function, since the arguments are also typed anyway.
`{ isApi }` instead of `isApi` seems to just be extra braces.

>This diff is plenty large
Yea I was trying not to change to much existing components and logic, especially  `SidePanel`, `Header` and `ApiPlayground` could still be improved a bit. I think we should keep that to seperate PR's tho.

I'll add the suggested change.

client/src/context/MDXContentContext.tsx
NicoZweifel: Oh yea, nice catch 👍 , I'll add the type definition to the declaration.

benjamin-shen:
> Would also be eager to hear your opinions about removing props from SidePanel, Header and ApiPlayground and using the context instead.

Since the context is serving as a shared/global store, my thought is that anything that we would normally put in a global store we should put in the context. So definitely not all props should be removed, just the ones that may be used across components horizontally. Specifically in the case of Api => ApiPlayground (in the demo PR), it feels more like normal vertical sharing. It definitely looks cleaner! But I'm not sure there is much benefit other than that. What are your thoughts?

Also, I was wondering why you went with this route instead of Redux and it looks like the Context API is [a lot better for our use case](https://dev.to/ruppysuppy/redux-vs-context-api-when-to-use-them-4k3p) (static data, smaller bundle size). I'm very happy you went this direction 😄

client/src/ui/MDXContentController/Footer.tsx
NicoZweifel: [This ](https://blog.logrocket.com/when-to-use-never-and-unknown-in-typescript-5e4d6c5799ad/)is a pretty nice post about `never` and `unknown`.


In this case it's just an alternative to `any` , that doesnt cause an error or eslint warning.
I used `never` because it got suggested by eslint, resolved eslint warnings.
It *should* never happen that href is undefined.

The Issue comes from `<Link>` requiring href of Type `Url`, but `previous.href` and `next.href` could be `string | undefined`.

Note that something like this would also work:
```tsx
{previous?.href && (
  <Link
    href={new URL(previous?.href)}
```
But I didn't want to make any additonal changes that could affect the logic of the component render inside of this PR.

benjamin-shen: Gotcha, very interesting!

NicoZweifel: the docs seem to have been depcrecated, [link ](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#the-never-type) to new docs

NicoZweifel:
> Also, I was wondering why you went with this route instead of Redux and it looks like the Context API is [a lot better for our use case](https://dev.to/ruppysuppy/redux-vs-context-api-when-to-use-them-4k3p) (static data, smaller bundle size)

Yea I agree, it is better for most use-cases I feel like nowadays.

Disadvantages like having state and ui in the same file don't really count as you can write custom hooks to encapsulate that.
With lots of dynamic data it also gets more interesting and a reducer and hook to encapsulate state becomes mandatory imo.

These issues are already covered tho with `useMDXContentController`, `useReducer` and the action types.

It basically comes down to a tradeoff:
-  Use redux, to potentially save code and time when you have a lot of complex different states all over the place.
This will require everybody to learn redux eventually and increases bundle size. 
-  Use your own reducers/context/hooks/architecture.

I feel like the tools, docs and general knowledge of the context api are getting better and better and there is less and less reason to use redux as time goes on.

client/src/context/MDXContentContext.tsx
NicoZweifel: Updated with suggested change.

client/src/hooks/useTableOfContents.ts
```diff
@@ -1,12 +1,17 @@
...
-export function useTableOfContents(tableOfContents: TableOfContentsSection[]) {
+export const useTableOfContents = (
+  tableOfContents: TableOfContentsSection[],
+  dispatch: Dispatch<MDXContentAction>
+) => {
   const [currentTableOfContentsSection, setCurrentSection] = useState(tableOfContents[0]?.slug);
   const [headings, setHeadings] = useState<any[]>([]);
 
-  const registerHeading = useCallback((id: string, top: string) => {
+  const registerHeading = useCallback((id: string, top: number) => {
```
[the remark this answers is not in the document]
NicoZweifel: Yea, I think so, as far as i remember it was complaining after extracting type definitions, that types don't match 😋

client/src/context/MDXContentContext.tsx
NicoZweifel: Updated with suggested change.

NicoZweifel:
@tobias-mintlify 
@benjamin-shen 
> Since the context is serving as a shared/global store, my thought is that anything that we would normally put in a global store we should put in the context. So definitely not all props should be removed, just the ones that may be used across components horizontally. Specifically in the case of Api => ApiPlayground (in the demo PR), it feels more like normal vertical sharing. It definitely looks cleaner! But I'm not sure there is much benefit other than that. What are your thoughts?

I agree about about props and not putting extra things in the context.
The context should only contain the state that is used for `MDXContentController`.

I improved the docs some more and closed the PR with `ApiPlayground`.

We could remove more props, and use the new context, like in `ApiPlayground`, specifically `SidePanel` and `Header`.
But I will leave it as is for now because it will get too large otherwise.

Thanks for your considerations, I think this is now good to go 🚀

docs/code/mdx-content-controller.mdx
```diff
@@ -0,0 +1,206 @@
...
+    const type = action.type;
+    switch (type) {
+        case MDXContentActionEnum.SET_STATE:
+        return {
+        ...state,
+        ...action.payload,
+    };
+    // other cases
+    ...
+    }
```
benjamin-shen, line 192:
```suggestion
    switch (type) {
        case MDXContentActionEnum.SET_STATE:
            return {
            ...state,
            ...action.payload,
        };
        // other cases
        ...
    }
```

benjamin-shen (commented):
A few doc nits

hahnbeelee (changes requested):
The footer doesn't have enough padding.
This branch:
image
Production:
image

tobias-mintlify:
I'm very impressed there was only one bug found 🚀

tobias-mintlify:
@hahnbeelee looks like this bug is unrelated to the PR, you can see the padding is messed up on [this page already](https://mintlify.com/docs/api-playground/demo). Looks like it's caused by API Playground pages. I'll look into fixing it but we can merge this in if that's the only bug you found.

tobias-mintlify:
I pushed a fix for the padding even tho it's unrelated to the PR.

client/src/ui/MDXContentController/SidePanel.tsx
```diff
@@ -0,0 +1,43 @@
+import { useMDXContent } from '@/hooks/useMDXContent';
+import { OpenApiResponseExample } from '@/layouts/ApiSupplemental';
+import { ContentSideLayout } from '@/layouts/ContentSideLayout';
+import { BlogContext } from '@/ui/Blog';
+import { TableOfContents } from '@/ui/MDXContentController/TableOfContents';
+
+export const SidePanel = () => {
```
hahnbeelee, line 7: Can we specify what side the `SidePanel` is on?

hahnbeelee: Maybe we should also rename `ContentSideLayout.tsx` it has always bothered me becuase it isn't clear which side it is.

hahnbeelee (approved):
Honestly haven't been able to test very in-depth because of how large this is - I wonder if there's a way for if we could've broken this refactor down. But overall, thank you so much for your work! I really really appreciate it. I can't wait to work off of this code :)

tobias-mintlify:
Han added a CLA bot this morning that blocks deploys, I'll merge after Nico signs.

hanywang2 (approved):

tobias-mintlify (approved):

benjamin-shen (approved):

client/src/ui/MDXContentController/SidePanel.tsx
NicoZweifel: Yea I agree that it was quite large and that it might be a bit hard to quickly wrap your head around the changes.

I tried to keep it to splitting state and the MDXContentController component for this PR
I am sorry if it was hard to get an overview.

The docs hopefully help a bit with that and understanding the concepts and changes.

> Can we specify what side the SidePanel is on?

We could change `SidePanel` and `ContentSideLayout` to have an `anchor` or `side` prop with `left | right` as type.

> Maybe we should also rename ContentSideLayout.tsx it has always bothered me becuase it isn't clear which side it is.

Yea, i think so too.
Renaming it or adding an `anchor` or `side` seem both good options.

docs/code/mdx-content-controller.mdx
```diff
@@ -0,0 +1,206 @@
...
+
+```tsx MDXContentController
+export function MDXContentController({ children, ...props }: MDXContentControllerProps) {
+    const ctx = useMDXContentController(props);
+    return (<MDXContentContext.Provider value={ctx}>
+            ...
+            {children}
+            ...
+        </MDXContentContext.Provider>
+    );
```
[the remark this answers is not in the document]
NicoZweifel: I formatted the doc files with prettier and pushed the changes.

PR #373, opened 2023-02-15
Clean up icon and navigation types

tobias-mintlify:


hahnbeelee:
Check these types out: https://github.com/mintlify/mint/blob/main/packages/mintlify/src/local-preview/listener/generate.ts#L12-L44

I thought long and hard about how to best name these types.

client/src/ui/Icon.tsx
```diff
@@ -27,6 +27,20 @@ export default function Icon({ icon, iconType, className, color }: IconProps) {
 }
 
 export function ComponentIcon({ icon, iconType, className, color, overrideColor }: IconProps) {
+  if (!iconType) {
+    return null;
+  }
+
+  // Validate the types
+  if (
+    !['brands', 'duotone', 'light', 'regular', 'sharp-solid', 'solid', 'thin'].includes(iconType)
```
hahnbeelee, line 36: Inside `config.ts` can you add and export
```
export const iconTypes = ['brands', 'duotone', 'light', 'regular', 'sharp-solid', 'solid', 'thin'] as const;
```
and change IconType to
```
type IconType = typeof iconTypes[number];
```

hahnbeelee (changes requested):

client/src/ui/Icon.tsx
tobias-mintlify: Trying this gave me type errors. That's saying IconType is an array of strings instead of "one of these strings".

hahnbeelee (approved):

PR #408, opened 2023-02-19
Bumping Next to 13.6

hanywang2:
## Summary

NextJS errors out and says that we need to set up `experimental-edge` instead of `edge`

CleanShot 2023-02-19 at 05 08 33@2x

## Test Plan

- Everything should build and run properly.
- Error should be gone


benjamin-shen:
This is a revert of https://github.com/mintlify/mint/pull/384

hahnbeelee (approved):

PR #410, opened 2023-02-20
Improve/Remove pop-in, render performance

NicoZweifel:
## Summary

Fixes pop-in on StickyCode blocks and request and response examples. Introduces a bunch of useCallbacks and Improves render performance.

## Test Plan

Tested locally with Mint docs and starter/empty custom docs.


hahnbeelee (approved):
Do you have any suspicions on where this might be happening elsewhere or do you think this covers everything?
Can you also explain to me what the change is? I know what bug this is fixing but I want to learn more about what was wrong initially and what you are changing now. You can leave comments on the PR to describe this maybe. I want to be more aware of what caused this so that I can keep a lookout and prevent this from happening in the future.

NicoZweifel:
> Do you have any suspicions on where this might be happening elsewhere or do you think this covers everything?
> Can you also explain to me what the change is? I know what bug this is fixing but I want to learn more about what was wrong initially and what you are changing now. You can leave comments on the PR to describe this maybe. I want to be more aware of what caused this so that I can keep a lookout and prevent this from happening in the future.

This should cover everything, I checked if everything is set correctly in the first render and it seems to be the case.

Yea the change is that instead of waiting and changing it in the hook after other dependencies are set (causing popin), the values are now set on first render immediately. This is similar to the old solution but achieved with useCallback to prevent generating/executing the functions in each render (like before the refactor). 

Let me know if you have more questions, we can also have a quick chat. 👍

PR #412, opened 2023-02-20
[WIP] Add link rot to CLI

hahnbeelee:


packages/mintlify-cli/src/index.ts
```diff
@@ -4,12 +4,16 @@
...
 
 export const cli = () =>
   yargs(hideBin(process.argv))
     .command(
       "dev",
-      "Runs Mintlify locally (Must run in directory with mint.json)",
+      "Run Mintlify locally",
       () => {},
       async (argv) => {
         await dev(argv);
```
NicoZweifel, line 19: I think this could be simplified to:
```suggestion
      dev
```
Beed to remove `}` manually, I was not able to insert a comment at line 20.

packages/mintlify-cli/src/index.ts
```diff
@@ -21,6 +25,25 @@ export const cli = () =>
...
+    .command(
+      "rename",
+      "Rename a file and update links in content",
+      () => {},
+      async (argv) => {
+        await renameFileAndUpdateLinksInContent(
+          argv.from as string,
+          argv.to as string
+        );
+      }
```
NicoZweifel, line 45: I think this can be shortened as well: 
```suggestion
     (argv) =>
      renameFileAndUpdateLinksInContent(
        argv.from as string,
        argv.to as string
      )
```

PR #417, opened 2023-02-22
Use PostHog internal analytics setup

hanywang2:
## Summary

This PR swaps out Segment for Posthog on internal analytics while fixing a few ESLint errors

## Test Plan
- No build errors
- Meant to be test in prod with proper configs


client/src/analytics/implementations/posthog.ts
```diff
@@ -8,8 +8,9 @@ import {
 
 export default class PostHogAnalytics extends AbstractAnalyticsImplementation {
   initialized = false;
+  subdomain?: string;
```
hahnbeelee, line 11: The fact that we had to add the `subdomain` property for post hog because we're using it internally and remove it from segment because we aren't using it internally seems like bad practice.
Analytics should be generalized and it's weird that we're editing them for our own purposes. At the least we should create a new class `InternalPostHogAnalytics` that extends `PostHogAnalytics`. We will most likely be refactoring analytics (given some of the eslint errors) so we can hold off on doing this for now.

hahnbeelee: Can you add a comment about it being for internal purposes or create the new internal object type?

client/src/analytics/internal/index.ts
```diff
@@ -1,8 +1,8 @@
-import { ConfigInterface } from '../AbstractAnalyticsImplementation';
-import SegmentAnalytics from '../implementations/segment';
+import { PostHogConfigInterface } from '../AbstractAnalyticsImplementation';
+import PostHogAnalytics from '../implementations/posthog';
 
-export default class InternalAnalytics extends SegmentAnalytics {
-  init(implementationConfig: ConfigInterface, subdomain: string) {
+export default class InternalAnalytics extends PostHogAnalytics {
```
hahnbeelee, line 4: Wait I made that comment before I saw this. You should def be editing this so that subdomain is a property here. Lmk if im wrong though/if this is difficult.

hahnbeelee (changes requested):

client/src/analytics/implementations/posthog.ts
hanywang2: Added a comment, but If you look at the way analytics is orchestrated, create eventListener creates a function and returns that. This means that the event data that actually needs to be sent is actually passed as a param within that function, meaning that you need the subdomain by that point. It's hard to explain, but tl;dr idk how else to do it

hahnbeelee (approved):
We will refactor this in the future

PR #419, opened 2023-02-23
Adding more events for logging

hanywang2:
## Summary

Add logging for the following events when a user 
- Clicks on the CTA
- Clicks on the navigation item
- Copies a code snippet (from the codeblock)
- Makes an API call

Note: Scrolling to the bottom of the page is NOT added in this PR

## Test Plan

Make sure to help test this one thoroughly. To do so, add the following code snippets to L135-136 of `AnalyticsMediator.ts`

```javascript
console.log(eventName);
console.log(eventProperties);
```

CleanShot 2023-02-05 at 16 42 47@2x

Then click around for the following events and you should see the above events get triggered properly.

**Note**: Make sure to double check that things aren't getting double logged


client/src/components/Accordion/Accordion.tsx
```diff
@@ -21,8 +22,8 @@ function Accordion({
   children: ReactNode;
 }) {
   const analyticsMediator = useContext(AnalyticsContext);
-  const trackOpen = analyticsMediator.createEventListener('accordion_open');
-  const trackClose = analyticsMediator.createEventListener('accordion_close');
+  const trackOpen = analyticsMediator.createEventListener(Event.AccordionOpen);
```
hahnbeelee, line 25: slay

client/src/enums/events.ts
```diff
@@ -0,0 +1,10 @@
+export enum Event {
```
hahnbeelee, line 1: slay

hahnbeelee (approved):
woot woot
tested all events. Another ez one we can add is card clicks
The only event that didn't work properly for me was Search but idk if it's weird locally.
CleanShot 2023-02-22 at 22 28 17@2x

PR #425, opened 2023-02-26
fix pop-in of api components

NicoZweifel:
## Summary
Fixes pop-in observed on https://heypal.mintlify.app/ask.
It was caused by a state update when `api` is undefined for a short duration causing it to reset.
The State updates now get prevented by checking if `api` is defined in the state's `pageMetaData` or `openApiPlaygroundProps`.

## Test Plan
Tested with the heyPal and mintlify docs.


client/src/context/MDXContentContext.tsx
```diff
@@ -84,6 +84,8 @@ export const MDXContentReducer: Reducer<MDXContentState, MDXContentAction> = (
   state: MDXContentState,
   action
 ) => {
+  // Uncomment for logging
+  // console.log(MDXContentActionEnum[action.type],action.payload)
```
hanywang2, line 88: Can you either
1. Explain a bit more on what this logs and why it's important to keep here or
2. Remove it?

hahnbeelee: Looks like it's helpful for debugging the main reducer.

NicoZweifel: Yea It's just helpful for debugging the reducer.

> Initially I was for keeping it to utilize it when debugging, but if we have good practices and are thorough with our PR descriptions we shouldn't need this comment.

pretty much this. Sometimes you might need it to debug state but then you can always put it back in.
I can add a section in the docs instead.

hahnbeelee (approved):

PR #430, opened 2023-02-27
Enable beta app directory in the next config file

tobias-mintlify:
## Summary

We are incrementally migrating to the `/app` directory as tracked in this issue: https://github.com/mintlify/mint/issues/429

This PR enables the `/app` directory and doesn't use any functionality yet. I am opening it so we can merge this small chunk separate from the migration work.

## Test Plan

`yarn dev-watch` and `yarn dev` work


client/app/layout.tsx
```diff
@@ -0,0 +1,7 @@
+// We are migrating to the /app directory incrementally.
```
hahnbeelee, line 1: What does this file do?

tobias-mintlify: [The Next 13 docs explain it best](https://beta.nextjs.org/docs/routing/pages-and-layouts#layouts). The summary is the root layout is like a merged `_app` and `_document` file. The long answer is every route can have a layout and Next.js uses them to avoid re-rendering the layout when a page changes within the same layout container.

hahnbeelee (approved):
- [x] `yarn build && yarn start`
- [x] `yarn dev`

Issue #432, opened 2023-02-28
Can't build previewing and mintlify-scraping


NicoZweifel:
**Describe the bug**
tsc exits with build erros.

**To Reproduce**
yarn && yarn build in `@mintlify/previewing`

**Expected behavior**
Successful build 

**Screenshots**
![image](https://user-images.githubusercontent.com/34443492/221874751-06d297d8-799e-4fa5-bab6-871eb3f9ac23.png)

I was able to fix the errors in previewing locally but in mintlify-scraping there are more errors:
`Found 22 errors in 11 files.`

**Device:**
 - OS: Win10

**Additional context**
Maybe this got introduced with recent changes to tsconfig? Or am I missing something that I need to setup?

PR #433, opened 2023-02-28
Previewing tsc

NicoZweifel:
## Summary
 I used this to fix the build errors in `previewing` https://github.com/mintlify/mint/issues/432

## Test Plan
Able to build `previewing` with these changes.


hahnbeelee (changes requested):
Ah apologies about the annoying build errors. Ben and I recently refactored the package structure for the CLI and when we ran it we swore it worked but I think it was a fluke.

Since this is also a recent refactor I'm not sure what the best way to test it is. I just went into `dist/local-preview/index.js` and commented  `await dev();` and navigated to our docs & ran the file (`cd ../../docs && node ../packages/mintlify-previewing/dist/local-preview/index.js`) but I got an error saying `Cannot read properties of undefined (reading 'port')`. Can you fix that too?

CleanShot 2023-02-28 at 11 19 26

I think it's line 155 at `src/local-preview/index.ts`

NicoZweifel:
> Ah apologies about the annoying build errors. Ben and I recently refactored the package structure for the CLI and when we ran it we swore it worked but I think it was a fluke.

Haha no problem, it probably didn't show because of local build files or sth.

> I think it's line 155 at src/local-preview/index.ts

I'll check but It was building fine for me with these changes. 
I was commenting dev and then using `node bin/index.js dev
` to test 😄. I was considering writing a test for a bit but was not sure if it was worth it yet since it would require quite a bit of mocking I guess.

NicoZweifel:
> I think it's line 155 at `src/local-preview/index.ts`

I changed to use optional chaining (?)
```ts
argv?.port as string
```

hahnbeelee (approved):

PR #436, opened 2023-02-28
check port

NicoZweifel:
## Summary
This is a work in progress and a proposal.
Checks if port is available and asks if next port should be checked recursively if not.

I would like to move this to `previewing` but am waiting for [this](https://github.com/mintlify/mint/issues/432) to be resolved.

## Test Plan

Tested by occupying ports locally and pressing `y/n` to launch/try on different port or return.


tobias-mintlify (commented):
Next.js automatically moves to the next port without prompting. Do you think it'd be safe to do the same here? Maybe we could check ~10 ports before asking if they want to keep searching for a port to avoid loops.

hahnbeelee:
@tobias-mintlify next automatically moves to the next port but `next-remote-watch` does not

hahnbeelee:
I actually think we can keep this in `mintlify-cli` I know it doesn't really make a lot of sense now but I think we should move all "cli functionality" out of `mintlify-previewing` so it's decoupled from the cli logic. So things like checking if yarn is installed or if the command is being ran in a folder where a mint.json file exists will happen in `mintlify-cli`

NicoZweifel:
> I actually think we can keep this in `mintlify-cli` I know it doesn't really make a lot of sense now but I think we should move all "cli functionality" out of `mintlify-previewing` so it's decoupled from the cli logic. So things like checking if yarn is installed or if the command is being ran in a folder where a mint.json file exists will happen in `mintlify-cli`

We could also write unit tests for a lot of this if it is decoupled.

NicoZweifel:
Added some tests and prettier. Ready for review 🚀

packages/mintlify-cli/src/cli.ts
```diff
@@ -0,0 +1,75 @@
...
+) => {
+  const port = (argv.port || 3000) as Parameters<typeof detect>[0];
+  const _port = await detect(port);
+
+  if (port == _port) {
+    return port;
+  }
+
+  const confirmed = await confirm(
+    `Port ${port} is already in use. Use port ${_port} instead? [Y/n]\n`
```
tobias-mintlify, line 31: Can we automatically look through other ports? Why do we need to ask the user if they want to keep looking? Other tools automatically switch so I'd like us to be the same if possible.

NicoZweifel: Yea we can switch automatically, afaik `webpack-dev-server` used to do it like this or still does, not sure. 
Next and vite do it automatically afaik.
I agree most of the time it is probably better to change it automatically but when it changes automatically and you don't expect it to (for example some node process didnt exit), it can be quite confusing.

I am not sure what is the better experience for customers that aren't as sophisticated with dev tools as I can imagine some people might be more comfortable with beeing reminded that there is already an instance running.

Does that make sense?
Other than that I don't see an issue with removing it, to me it would be fine, since as a user I'd check the port anyway, but I can see how some users might not and might think the cli is not updating or sth.

tobias-mintlify: Good point people might not realize another is already running

tobias-mintlify: The CLI is not designed to run multiple versions at once. They all use the same `mint` folder. So it's probably best to prevent them from running multiple. The prompt will be good for that.

tobias-mintlify (approved):

PR #441, opened 2023-03-01
Get allOf to work in api playground

hahnbeelee:
## Summary
Title
## Test Plan
Go to http://recall.localhost:3000/reference/bot_create and make sure the api playground types match up with https://recallai.readme.io/reference/bot_create

before

CleanShot 2023-02-28 at 22 55 13@2x

after

CleanShot 2023-02-28 at 22 54 53@2x


client/src/ui/MDXContentController/getOpenApiPlaygroundProps.tsx
```diff
@@ -79,6 +79,12 @@ export function getOpenApiPlaygroundProps(
   // Get the Body ApiComponents
   if (bodySchema?.properties) {
     Object.entries(bodySchema.properties)?.forEach(([property, propertyValue]: any, i: number) => {
+      if (
+        propertyValue?.allOf &&
+        Array.isArray(propertyValue.allOf) &&
```
tobias-mintlify, line 84: `Array.isArray()` returns false when you pass undefined so you could have also shortened this to `Array.isArray(propertyValue?.allOf) && propertyValue.allOf.length > 0`

tobias-mintlify (approved):

PR #444, opened 2023-03-01
Add noindex meta tags on hidden pages

tobias-mintlify:
## Summary

Hidden pages should not be indexed. We need to set `<meta name="robots" content="noindex">` on them.

Originally, we had planned to also disallow hidden pages in robots.txt. You can see that in this closed PR: https://github.com/mintlify/mint/pull/443 However, [Google advises to allow crawling hidden pages](https://developers.google.com/search/docs/crawling-indexing/block-indexing) so they can see the `noindex` instruction. Otherwise, they may include them in results if another site linked to the page but they did not crawl the page themselves.

## Test Plan

Visit a hidden page like http://mintlify.localhost:3000/code-of-conduct#2-warning and verify `<meta name="robots" content="noindex">` is set in the <head>.


client/src/hooks/useCurrentPath.ts
```diff
@@ -9,11 +9,11 @@ import { ConfigContext } from '@/context/ConfigContext';
 export function useCurrentPath() {
   const router = useRouter();
   const { subdomain } = useContext(ConfigContext);
-  const withOutBasePathRemoval = '/_sites/' + subdomain;
-  const withBasePathRemoval = router.basePath + withOutBasePathRemoval;
+  const withoutBasePathRemoval = '/_sites/' + subdomain;
```
hahnbeelee, line 12: nice loll

client/src/pages/_sites/[subdomain]/[[...slug]].tsx
```diff
@@ -119,6 +120,10 @@ export const getStaticProps: GetStaticProps<PageProps, PathProps> = async ({ par
       console.log(`⚠️ Warning: MDX failed to parse page ${path}: `, err);
     }
 
+    // When we move to the /app directory this should go in the metadata function. We use the hidden pages
+    // to generate a noindex meta tag for pages that are hidden.
```
Comment (hahnbeelee) [inline client/src/pages/_sites/[subdomain]/[[...slug]].tsx:124]: Can you change this comment to be more like a TODO? like

```
TODO - move this to metadata function when we migrate to /app directory
```
Does the statement "We use the hidden pages to generate a noindex meta tag for pages that are hidden." Apply to now or in the future?

client/src/ui/CanonicalTags.tsx
```diff
@@ -4,20 +4,25 @@ import { useState, useEffect } from 'react';
...
 
   useEffect(() => {
     setOrigin(window.location.origin);
   }, []);
 
-  const currentUrl = origin + router.basePath + useCurrentPath();
+  const currentPath = useCurrentPath();
+  const currentUrl = origin + router.basePath + currentPath;
+
+  // Remove leading slash before comparison
```
hahnbeelee, line 18: Although I don't believe in the argument that code should be self-documenting so you shouldn't write any comments. I do think in this case you can just do:
```
const pathWithRemovedLeadingSlash = currentPath.slice(1);
const isHidden = hiddenPages?.includes(pathWithRemovedLeadingSlash);
```
The comment really feels like overkill in terms of context its giving so I feel like you should just change it to the above or remove it.

hahnbeelee (changes requested):
Is there a reason why we're fetching the hidden pages array instead of sending the subdomain and the path as an input to the backend?

client/src/pages/_sites/[subdomain]/[[...slug]].tsx
Comment (tobias-mintlify) [reply client/src/pages/_sites/[subdomain]/[[...slug]].tsx]: Updated the comments

hahnbeelee (approved):
🚢

PR #445, opened 2023-03-01
Revert Next.js to 13.2.0 until the link bug is resolved

tobias-mintlify:
We need 13.2.0+ to support routes that we use for sitemaps but any newer versions introduced a bug in link routing. We are reverting until the bug is fixed.


hahnbeelee (approved):
Tested:
- Link bug fixed in development (at subdirectory path)
- sitemaps still work
- robots.txt still work

PR #446, opened 2023-03-04
Remove local env variables

benjamin-shen:
## Summary
Added default values and refactored `process.env` such that local env variables are unnecessary for single-tenant hosting (i.e. for open-source contributors that want to make changes to client). `IS_MULTI_TENANT=true` and `ADMIN_TOKEN` are still required for multi-tenant hosting.

After this PR is merged, devs will be able to simplify their local `.env` files and there will be reduced friction to start contributing to `client/` :)

## Test Plan
1. Go into the `client/` folder.
2. Comment out your entire `.env` file.
3. Run `yarn preconfigure`, then `yarn dev`. Navigate to `http://localhost:3000` and sanity check.
4. Add back `IS_MULTI_TENANT=true` and `ADMIN_TOKEN` (but not `API_ENDPOINT`) to your `.env` file.
5. Start server locally. Navigate to `http://mintlify.localhost:3000` and sanity check. Validate `sitemap.xml` and `robots.txt`.

## TODO
- [ ] Reorganize/remove `lib/` folder and remove exception from `.gitignore`
- [ ] Clean up Infisical secrets in Mintlify Client
- [ ] Clean up Vercel env variables
- [ ] Add instructions (internal docs) for multi-tenant local hosting


client/README.md
```diff
@@ -21,9 +21,9 @@ yarn
 
 ### Local Development
 
-- Log in to Infiscal for the .env variables. The project is `Mintlify Client`. Save those env vars to `client/.env`.
+- Run `yarn preconfigure` or `yarn preconfigure <path/to/docs>`
 - Run `yarn dev`
-- Navigate to `http://mintlify.localhost:3000/`
+- Navigate to `http://localhost:3000/`
```
hahnbeelee, line 26: This makes 100% sense for open source contributors but the old version helps a ton for internal devs. Maybe we should just write internal docs at least specifically for mint. Too many different stakeholders are looking at this repo.

client/src/constants.ts
```diff
@@ -0,0 +1,10 @@
+export const API_ENDPOINT = process.env.API_ENDPOINT || 'http://localhost:5000';
```
hahnbeelee, line 1: this file is chef's kiss

client/src/data-fetching/getHiddenPages.ts
```diff
@@ -1,10 +1,4 @@
...
-      headers: { Authorization: `Bearer ${process.env.ADMIN_TOKEN}` },
-    }
-  );
-  const hiddenPages = await hiddenPagesResponse.json();
-  return hiddenPages;
-};
+import { fetchWithAuth } from './fetch';
+
+export const getHiddenPages = async (subdomain: string): Promise<string[]> =>
+  (await fetchWithAuth(subdomain, '/hidden-pages')).json();
```
hahnbeelee, line 4: dis is so clean

hahnbeelee:
What about INTERNAL_ANALYTICS_WRITE_KEY?

hahnbeelee (approved):
- [x] Successful test plan

We gotta improve internal documentation

client/README.md
hahnbeelee: Oh I understand - this is your last todo list item

benjamin-shen: The old instructions also were incomplete on the backend side. We should keep all the public documentation relevant to open-source

benjamin-shen:
> What about `INTERNAL_ANALYTICS_WRITE_KEY`?

I believe that's only used in production, right? No need for local development

client/README.md
hahnbeelee: sounds good

hahnbeelee:
> I believe that's only used in production, right? No need for local development

Yeah that's true

hahnbeelee:
> I believe that's only used in production, right? No need for local development

Wait actually I don't really view this as "removing" local env variables, I kind of just view it as a refactor to consolidate all the places we do use an environment variable to make the code cleaner. All the environment variables in `constants.ts` are used for enterprise mintlify features (not self-hostable)

hahnbeelee (approved):
Did some white box testing. LGTM

PR #448, opened 2023-03-06
Lowercase file extensions

benjamin-shen:
## Summary
Fixes the following issue by lowercasing file extensions:
image

No need to release immediately unless someone on the migration team encounters the issue.

## Test Plan
Couldn't reproduce the original issue (due to lack of link) 😢 but the scraper still seems to scrape other sites fine.

image


tobias-mintlify (approved):
You also need to update the server. There's a similar function there.

Issue #326, opened 2023-02-07
Auto-generated requests examples for the API Playground don't encode the basic authentication headers


tobias-mintlify:
We request the two authentication values and pass them as headers when they should be base64 encoded. Contributor will need to update the code at `client/src/utils/apiExampleGeneration/generateApiRequestExamples.tsx`. You will likely need to add a parameter to the function to receive the type of authentication so you can extract the authentication headers needed for the encoding.

For example:

`-H "Authorization: Basic <base 64 encoded client_id:api_key>"`
or
`-u "client_id:api_key"`

You can see this on [Dot's website](https://docs.dots.dev/api-reference/v2/users/get-users)

PR #451, opened 2023-03-06
Use base 64 encoding when generating basic authentication headers

tobias-mintlify:
## Summary

Fixes https://github.com/mintlify/mint/issues/326

Basic authentication is supposed to base 64 encode requests. We were wrongly generating separate header entries for each value.

## Test Plan

dotsdev.localhost:3000 is working correctly

Before each generated example had a header for each value:
Screen Shot 2023-03-06 at 3 01 38 PM
Now we encode them:
Screen Shot 2023-03-06 at 3 14 14 PM


hahnbeelee:
> dotsdev.localhost:3000 is working correctly

What does correctness mean? Both the issue and the PR description don't give me pointers on what behavior I should be expecting from this PR. Can you be more descriptive?

I can derive the desired behavior by looking at the test cases and digging into the code but it's much easier if you provide screenshots of the before or after or describe how I can verify correctness.

tobias-mintlify:
@hahnbeelee updated the test plan

hahnbeelee:
Is the intended behavior that it only works if both of them are defined?

CleanShot 2023-03-06 at 16 50 00

hahnbeelee:
Thoughts on making the default value `<INSERT_AUTH_VALUE>` instead of `AUTH_VALUE` to make it explicit that you should be updating it?

CleanShot 2023-03-06 at 16 51 19

client/src/layouts/ApiSupplemental.tsx
```diff
@@ -90,11 +93,15 @@ export function GeneratedRequestExamples({
   paramGroupDict,
   apiPlaygroundInputs,
   apiBaseIndex,
+  authMethod,
```
hahnbeelee, line 96: It kinda feels like `generateRequestExample` and `GeneratedRequestExamples` can be merged into 1. The only place `GeneratedRequestExamples` is used is in [useApiPlaygroundCallback](https://github.com/mintlify/mint/blob/main/client/src/hooks/useApiPlaygroundCallback.tsx#L37-L42) and you have access to `ConfigContext` there.

hahnbeelee (approved):
Preferably can you refactor `generateRequestExamples` to be `GeneratedRequestExamples`? And if you want to hold off on doing it can you write a comment to do so in the code.

client/src/layouts/ApiSupplemental.tsx
tobias-mintlify: Left a comment about moving the `ConfigContext` dependency out in a future PR. When we refactor this I think the `generateRequestExample` should just return a javascript object instead of generating the full `<Code>` component. That would let us unit test the generation logic like I originally intended to.

tobias-mintlify:
I'll refactor this on Monday

PR #452, opened 2023-03-06
Fix modeToggle.default='dark' overriding user preferences

tobias-mintlify:
## Bug Description

1. `modeToggle.default` is set to `dark`
2. User toggles to light. `localStorage.isDarkMode === "false"`
3. Hard-refresh the page. Page is now dark.

The bug is caused by the loading script never checking if the user preferred light mode when the default is dark. We only checked if the user preferred dark mode and the default was light.

## Test Plan

Visit color-test.localhost:3000

Toggle to light.

Hard-refresh, the page should still be light.

## Style Notes

I added more if-statements to the script even though or-statements are also valid. I think this is easier to read but like everything I expect it'll keep changing over time.


hahnbeelee (changes requested):
The test plan didn't work for me. Yes I'm on latest and on the branch

https://user-images.githubusercontent.com/55263191/223293474-6f6e8799-f67e-41db-8ee6-a452589f8881.mp4

hahnbeelee (approved):
I think this is a better UX although one edge case this doesn't account for is auto appearance switching:

CleanShot 2023-03-09 at 12 42 29@2x

PR #453, opened 2023-03-06
Prevent scrollbars causing layout shifts

tobias-mintlify:
## Summary

Two types of dealing with scrollbar layouts shifts:

1. Page scrollbar will not overlay instead of shifting. If we need the old padding we can add extra padding to the right of the page equal to the scrollbar's width. This PR reduces right padding by a few pixels on most pages because the scrollbar will not be adding spacing anymore. This is not noticeable so we can revisit alternate designs whenever we update page paddings next.
2. Navigation will now have stable scrollbar gutters. The space is always reserved whether the scrollbar is there or not. This preserves the current padding. This also means there is slightly more right padding than left padding because the scrollbar adds space.

## Test Plan

Run `yarn prebuild` then `yarn dev-watch` and navigate between http://localhost:3000/components/code-group and http://localhost:3000/components/tooltip


client/src/css/base.css
```diff
@@ -7,6 +7,7 @@
 html {
   height: 100%;
   font-feature-settings: 'cv02', 'cv03', 'cv04', 'cv11';
+  overflow-y: overlay; /* Shows scrollbar over content instead of causing a layout shift */
```
hahnbeelee, line 10: Why is the implementation different for the main page and the nav bar?

tobias-mintlify: I couldn't get scrollbar gutters working on the html tag. I read a blog post claiming browsers treat it as a special case but I couldn't find a better source.

hahnbeelee (approved):

PR #457, opened 2023-03-07
Update shouldDeploy script for client

benjamin-shen:
## Summary
Updates script from #197. I thought it was weird that one commit would fail the deployment while the next commit (that didn't touch `client/`) would appear successful.

Now, the script should cause Vercel to deploy if there is any client diff from main, not just a diff from the previous commit.

## Test Plan
Make changes in `client/` and make sure `git diff --name-only main -- .` detects them.

## Note
We should make sure this continues to deploy on commits to the main branch.


tobias-mintlify (approved):
I'll monitor this after you merge.

PR #458, opened 2023-03-07
Get single tenancy to deploy

hahnbeelee:
## Summary
I tried getting single tenancy to deploy for preview deployments and now also for password protected docs. Although the build was outputting that it was building correctly everything was returning a 404. I realized this was happening because the middleware was gating single tenancy from deploying.

### To-Do
- [ ] Fix redirects. I have been looking into this for far too long. I'm just going to move on for now. It works with `yarn build && yarn start` locally. idk why the redirects aren't working when deployed. For example: https://preview.mintlify.com/ isn't deploying. Anyone know how to debug?

## Test Plan
- Make sure code makes sense
- Check out https://preview.mintlify.com/quickstart


client/src/middleware.ts
```diff
@@ -23,8 +23,7 @@ export default function middleware(req: NextRequest) {
     : hostname.replace(/\.localhost:\d{4}/, '');
 
   // rewrite root application to main folder
-  if (hostname.match(/^localhost:\d{4}$/)) {
-    // TODO: change so it detects if it's at a subdomain or not
+  if (hostname.match(/^localhost:\d{4}$/) || !IS_MULTI_TENANT) {
```
hahnbeelee, line 26: Idk if there is a better way to do this.
This is what the `url` value looks like in multi-tenancy. I'm sure we can break apart the search param and look for whether there's a subdomain set but I feel like it can have unknown repercussions.

CleanShot 2023-03-06 at 23 33 43

tobias-mintlify (commented):
Are you sure you meant to put `|| !IS_MULTI_TENANT` instead of `&& !IS_MULTI_TENANT`?

I'm wondering why a multi-tenant deployment would want to leave `localhost:3000` as-is instead of going to the next return statement with
```
url.pathname = `/_sites/${currentHost}${url.pathname}`;
```

hahnbeelee:
@tobias-mintlify I think OR is fine because it gives us the flexibility of having both single tenancy and multi tenancy on at the same time in development. Preferably we can also have both in production but I think that's a separate ticket that will require a lot of testing.

tobias-mintlify (approved):

PR #459, opened 2023-03-07
Rename Koala publicId to publicApiKey

tobias-mintlify:
## Summary

Koala is changing their property name to `publicApiKey`

## Test Plan

Unit tests should pass.


hanywang2 (approved):
Thanks for getting on this so quickly!

